PR #113, opened 2018-04-04
Fix Vichy France not being created properly

Incnas:
Also fixes French surrender event to add British-Franco union. Done by migrating code from base game.


Greatexperiment:
Remove is no longer used in the game, so adding it back in creates problems. There are also some redundant event IDs.

Thinking-waffle:
I made changes it can be closed.

Issue #111, opened 2018-04-01
France ban communism through an event and still has the decision (note probably same for Belgium)


Thinking-waffle:

Greatexperiment:
This issue has now been partially addressed, but it's up to you if you want to close it.

Zurich899:
I just made a pull request that will stop France from getting the ban communism event if it bans communism through a decision (by adding a minimum 5% communism requirement to the trigger) and it seems to work in testing. I think that this issue can be closed if the pull request is accepted, since the decision and event for France are mutually exclusive, and Belgium's relationship to communism is being handled as part of Waffles' Belgium overhaul.

Tyguy9112:
Reply, if issue reoccurs.

PR #114, opened 2018-04-06
Issue #111 Fix - Disables "Ban Communism" and "Anti-Communist Raids" if France bans communism through event.

Zurich899:
Fixes Issue #111 by preventing the ban communism and anti-communist raid decisions from appearing if France has banned communism through an event. Unfortunately, I could not find the event that bans communism in Belgium (but I could add a similar condition if someone could point it out to me)

### DIFF
--- a/common/decisions/political_decisions.txt
+++ b/common/decisions/political_decisions.txt
@@ -1651,11 +1651,17 @@ political_actions = {
 
     	visible = {
 			NOT = { has_government = communism }
+			NOT = { 
+				AND = {
+					tag = FRA
+					has_idea = FRA_communism_banned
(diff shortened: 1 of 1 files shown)


Thinking-waffle:
Due to the anti German nature of the Belgian security response, only fascism can be banned through an event (and at the same time Van Severen sent a letter to the king to accept his equivalent of the blackshirts into the army to protect the country)

the event is tblra.131 and doesn't include Van Severen's proposal in it's RT56 version as I coded that quite recently and I had ideas to make its effect more dynamic than a simple flat bonus.
Thanks for your contribution.

Issue #100, opened 2018-03-30
The formation of vichy puppet free france


Thinking-waffle:

Zurich899:
After a bit of research, I learned that Vichy Free forms through a civil war, and that the civil war which creates Vichy was removed from R56's event file (no idea why). I did some testing, and adding the civil war command back to the file makes Vichy work correctly.

Greatexperiment:
It's because it was causing errors and CTD issues due to some modded content that Waffle needs to update. So it was temporarily removed until it's fixed. However, this removal wasn't very graceful and was meant to just be a quick fix.

PR #128, opened 2018-04-13
#100 Fixing Vichy France

Zurich899:
Adds code from the basegame which creates Vichy France through a Civil War if France is defeated by Germany. This code was removed from the R56 France event file for some reason, and it was preventing Vichy France from being formed (with Free France maintaining ownership of southern France)


Greatexperiment:
Waffle should review this as it has to do with his mods.

PR #135, opened 2018-04-13
Estonia fixes and asset touch-up  

SpicyAlfredo:
**Bug Fixes**
1. Fixes _EST_air_innovation_ not building airbases in Estonia
 > Now Estonia workers will build airbases in Estonia. Prioritizing states (191) and (13). Instead of choosing a random place on the planet earth to build. 
2. Fixes _EST_air_innovations_ tool-tip being incorrect about airbase location.
 > This is a incomplete fix to the issue. For example if you don't own state (191), (13) it will incorrectly say the airbases where built there. 
3. Fix bad reference for decide planner trait GFX
 >You can go into battle lacking important intelligence even faster! 

**Flavor Text**
1. Adds description for _EST_help_from_allies_
> We alone cannot defend Estonia from it's enemies. We must seek outside help from the west.
2. Adds description for _EST_special_marine_forces_
> "Any offensive operations in the Baltics will require a amphibious force. We must reestablish the 'Meredessantpataljon' "

**Portraits** 
1. Removed portraits with no references
 > Someone really hates fast load times
2. Updated Johan Laidoner with a better art asset
 > He lost weight
3. Updated the leader of Estonia with the new base game asset
 > Looks much better
4. Add portrait for Gustav Jonson
 > further info below 
5. Removed portrait for Aleksander Einseln
 >further info below 

**History** 
1. Removed Aleksander Einseln as a General 
 > He would had been six years old in 1936...
2. Added Gustav Jonson to replace Aleksander Einseln as a General 
 > Level 2 skill, Traits: Cavalry officer, Winter specialist. 
3. Johan Laidoner Homeland Connoisseur trait removed, added War Hero.
 > Changed Traits makes more historical sense for Ladioner.
4. Added a certain Italian General who you shouldn't promote to Marshal of Italy 
 >Don't let him near Libya...


Greatexperiment:
Thanks for your work.

PR #158, opened 2018-04-18
Issue #143 Found the placeholders

Zurich899:
It looks like Paradox added new Sprite types for placeholders in one of the files that R56 overrides. Adding the code from the base game removes that placeholder message from the error log.


Greatexperiment:
Hooray!

PR #176, opened 2018-04-24
Fixed Ethiopian Infantry Expert having "Not: Purged By Stalin" requir…

Plutozz:
…ement.

The Ethiopian infantry expert "Wehib Pasha" unlocked through the The Ferenghi focus also had a "Not: Purged by Stalin" requirement which obviously makes no sense - so I got rid of it. Simple. (I even researched and he was not related to the Soviet Union in any way)


Rokim21:
Good bugfix!

PR #178, opened 2018-04-25
Issue #65 Added text and desc. for Belgian Tank Destroyer focus and i…

Plutozz:
…dea.

Added text and a description for the tblra_continue_the_tank_destroyer_program focus for Belgium along with the same for the national spirit it provides.


Rokim21:
Include that already.

PR #179, opened 2018-04-25
Some Tweaks

Rokim21:
Generals can now lead 30 divisions, Field Marshals can lead 100 and army,navy and airforce experience is capped at 1000.


Greatexperiment:
Why is this needed?

Rokim21:
Why not?

Greatexperiment:
If there's a case to be made, sure. It needs to improve gameplay, or be a much requested feature.

Rokim21:
It improves gameplay, it's too easy to get at 500 xp which makes upgrading equipments kinda shitty, also a 30 divisions general is much better for small nations, and a 100 divisions field marshal makes sense and it's useful for massive countries like the Soviet Union and the German Reich.

Greatexperiment:
The xp cap increase is fine. Not going to mess with the general / marshall caps though.

Rokim21:
Then that way it will be!

Issue #173, opened 2018-04-23
Invalid Scope


Greatexperiment:
[12:37:00][effect.cpp:411]: Error in effect:set_country_flag: Invalid Scope
[12:37:00][trigger.cpp:242]: Error in trigger:: Invalid Scope

This is likely from one of the recent additions. It's also going to be a pain to find.

Zurich899:
Odd, I don't have those lines in my error log when loading the beta. I just have a lot of text about missing ideas (which are from the War at Sea mod, I think).
[error.log](https://github.com/deliciousmods/1956_beta/files/1949607/error.log)

Greatexperiment:
There shouldn't be any errors associated with the TWAS stuff - even if it is incomplete at the moment. The scope error appears when the game is fully loaded after selecting a country.

Zurich899:
Ah, I see what you mean. After I loaded a country I saw it too. I'll try looking for it by working through the latest commits, but you're right, it's definitely a needle in a haystack.

Zurich899:
This really is a pain to find. But I can confirm it's not in the TWAS, new France, or Honduras event files, since I tried deleting them in my local beta and still got that error.

Zurich899:
And it's not in any of the event files. I deleted literally everything in the events folder but still got that error. What uses triggers, other than events?

Greatexperiment:
🥇

PR #189, opened 2018-04-28
Issue #173 Found the needle in the haystack

Zurich899:
After looking for the needle in the haystack, I decided to burn the hay and pull the needle out of the ashes. After deleting almost every other file in the game, I finally narrowed it down to this file, and found that there was an if statement being used incorrectly (which caused the bad trigger). ROOT also should have been every country, which caused the bad scope.

### DIFF
--- a/common/on_actions/00_dnm_on_actions.txt
+++ b/common/on_actions/00_dnm_on_actions.txt
@@ -1,20 +1,15 @@
 on_actions = {
 	on_startup = { 
 		effect = {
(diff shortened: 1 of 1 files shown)


Greatexperiment:
It's nice to see that it wasn't anyone's fault - something from TWAS. Thanks for dealing with this. Normally I'm the unfortunate soul for this sort of job. Glad to share the misery. 🗡

PR #193, opened 2018-04-29
Over 14 GFX additions. 13 focus icons and 1 portrait.

Plutozz:
Over 14 new GFX additions including 13 focus icons: 5 for NOR, 1 for SOV, 1 for IRE, 1 for JAP, 1 for EST and 4 for all. Along with a portrait of Albert Sarraut for FRA.

I have also updated the Yaroslavl SOV state - changing the Synthetic Refinery into a Rubber Refinery which is what it was historically.


Plutozz:
Made use of some the added GFX into 4 focus trees. More to come at a later date.

PR #208, opened 2018-05-02
Fixes to SAU tree, with slight buff

SpicyAlfredo:
Fixes bug also slight buff to starting ideas, and ability to remove nomads idea though army tree. Always having it seems like a handicap late game


Thinking-waffle:
Indyclone will make a new saudi tree
with research and all. The layer is already set up but he was a bit busy recently he should restart all of this pretty soon.

SpicyAlfredo:
This is just a band aide to a major CTD issue with the current tree that should be pushed asap, regardless of a new tree coming on the way

PR #209, opened 2018-05-03
Delete honduras.txt

Zurich899:
When I uploaded the Honduras national focus file with the sanity scripts, Github seems to have uploaded a second national focus file instead of replacing it. We should delete the older version of this file.


Thinking-waffle:
We could make original suggestions based on updated content.

Zurich899:
What do you mean by that? This file only exists because of a Github error. The updated (and correct) national focus file is https://github.com/deliciousmods/1956_beta/blob/Beta-Edits/common/national_focus/Honduras.txt (the capitalized national focus file, which Github, for some reason, interpreted as different from the not capitalized file)

PR #216, opened 2018-05-04
US flavor

SpicyAlfredo:
I was bored. Also Commanders traits levels adjusted or added


SpicyAlfredo:
With the new Unpopular trait!

PR #230, opened 2018-05-07
Minor fixes

SpicyAlfredo:


Thinking-waffle:
The fix for wallonia is wrong actually and should not be added. I did it the correct way the other day

SpicyAlfredo:
Changed to be current WLL file

PR #237, opened 2018-05-13
Fixing a CTD caused by forming Vichy France

Zurich899:
Someone who plays R56 was nice enough to upload his save game, and I found the source of his crash. The event where France surrenders and becomes Vichy was trying to remove a leader that France didn't have. Adding checks to all the leader removals seems to fix this CTD (although the game lags for a few seconds when Vichy is formed)

### DIFF
--- a/events/France.txt
+++ b/events/France.txt
@@ -1471,15 +1471,54 @@ country_event = {
 				}				
 			}		
 			##add checks if general exists
-			remove_unit_leader = 801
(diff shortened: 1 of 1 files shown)


Greatexperiment:
The changes here are such that we can't merge straight into the branch. The file has to be manually merged on the code level. This could be for a number of reasons.

Thinking-waffle:
I have done those changes alreay

PR #254, opened 2018-05-24
bad pull request

SpicyAlfredo:


SpicyAlfredo:
Ready to Commit to Master. GXP should do a glance over the files to catch any glaring errors I might had missed. Germany OOB and tiny adjustments to snowballing of Germany PP gain. Also with the removal of our KMT Jap content. We get the the new 1.5 JAP content as a result.

Greatexperiment:
I might have missed this conversation, but why are we pulling KMT Japan content?

Greatexperiment:
Also, I will review this - but given the number of changes it will take a while. I will likely play through it before approving the merge as well.

Greatexperiment:
Should this one be closed?

PR #263, opened 2018-05-31
Added and integrated two new focus icons.

Plutozz:
First one being the Italian Cooperation icon - integrated into the Belgian Congo and Yugoslavian trees. Second one being the China Mongolia icon being implemented into N. China!


Rokim21:
That Chinese icon was supossed to replace GFX_goal_kmt_china on the focus KMT_internal_stability

PR #283, opened 2018-06-13
More Compatible Color,a Argetinian Portrait and some flags

Rokim21:


Greatexperiment:
This can't be merged, as a number of files need to be manually edited to fit. Please make adjustments according to the list.

PR #331, opened 2018-07-29
Quality of life changes.

Plutozz:
Added Siberia impassable zone, added some space and tiny tweaks into Iran's focus tree and some minor Brazil tweaks. Updated bookmarks.


SpicyAlfredo:
I don't think we're adding new states until the reboot

Plutozz:
It's not new states though. It's simply just a added impassable zone. Just like what we did to Africa - unlike this time it's added instead. I understand the concern though.

Plutozz:
I can get rid of it if need be.

SpicyAlfredo:
Send me a screenshot of what it looks like

PR #340, opened 2018-08-06
MattLongCT UKR Patch adjustments

MattLongCT:
1. Add Andriy Livytskyi as Leader  of UKR (oligarchy)
2. Give UKR a claim on 73-Carpathian Ruthenia


SpicyAlfredo:
Thanks for the Work! Maybe Ukraine can get the core when Carpathia spawns?

PR #456, opened 2018-11-30
Sven december changes for greece

sventhedanishcitizen:


SpicyAlfredo:
Your field Marshalls needs to be offensive, because he's a chief of army for offensive so one has to be switched since he can't be both

PR #479, opened 2018-12-14
yugo

Darkneon312:
Rupnik like ( maginot line for yugoslavia after anschluss and italian claims)


Greatexperiment:
Seems good. Thanks!

PR #480, opened 2018-12-14
Waffle return

Thinking-waffle:
bug fixes for previous content I have made, mainly Croatia but also Armenia, Luxembourg, the Netherlands, Belgium. In addition to that, I added 1 oil in Estonia.


Greatexperiment:
Welcome back!

Reviewed aaaaand... seems okay.

PR #481, opened 2018-12-14
changes requested by waffle i forgot to pull request

Darkneon312:
yea... i forgot again: combat reports xp halved, oil in washington usa, oil in tartu estonia


Greatexperiment:
Currently there's an unresolvable conflict with history/states/191-Tartu.txt

Thinking-waffle:
I don't think that it's really a conflict, we just want the same thing

PR #500, opened 2019-01-23
Reworked the First Vienna Award.

Plutozz:
- Reworked the First Vienna Award so that it's now done through events following the Fate of Czechoslovakia focus instead. (Requested by Waffle)
- Updated Austrian focus icons.
- Germany now starts with the Prepare Invasion of Poland focus completed in 1939.
- Minor 1939 fixes. (there's still some minor errors there but they've been there for a while at this point. I'll get around to fixing them soon unless someone else gets to it.)


Greatexperiment:
Error Log is clean. Be sure to update the wiki notes.

PR #508, opened 2019-01-31
Fixes for Swedish fascist and German focus trees.

Plutozz:
* Added options to the Icelandic player to fight back the Swedish supplied mutiny. This might lead with Sweden taking direct action though!
* Germany should now have done roughly 20 focuses or so in the 1939 start.
* Fixed the Icelandic Mutiny focuses for Sweden always being bypassable.
* Fixed War in France focus for Germany not being able to be bypassed if France doesn't exist.


Greatexperiment:
Looks good. Be sure to update the wiki notes.

PR #509, opened 2019-02-01
Waffle return

Thinking-waffle:


Greatexperiment:
Seems fine. Update the patch notes.

PR #515, opened 2019-02-03
fix portrait Latvia and Austria (using dlc portrait for DoD owners)

Thinking-waffle:
fixing portraits for Austria and Latvia


Thinking-waffle:
I chose to remove the file, i just hope that it won't replace it with nothing.

PR #520, opened 2019-02-04
Fixed space error in localisaton files about Albania.

ysjbserver:
Fixed space errors before exclamation mark.


Greatexperiment:
It's one of those small things, but it's good that you noticed it. Be sure to always update the wiki notes with your changes.

PR #534, opened 2019-03-05
MTG Fixes

PordyHUN:
Fixed conflicting Polish and Brazilian states, because of the vanilla map edits.
Removed the OIL cost from equipment productions,
Added 21 AIR WING XP level, like the army and navy,
Added fuel consumption values for air units.
Fixed numerous missing portraits, and leader icons.


Greatexperiment:
Great work!

PR #538, opened 2019-03-07
Yugo to Soviet tech tree bonuses fixed

PordyHUN:
Other minor issues fixed.


Greatexperiment:
Conflicts with some of waffle's edits.

Thinking-waffle:
I had to move some focuses around because the error log don't like when a relative position ID is pointing to a focus that has not been written yet.

PR #551, opened 2019-03-20
Small fixes for the greater good!

PordyHUN:


Plutozz:
I'd be really careful with this for now since it seems to revert several changes that was made a while ago.

PR #574, opened 2019-04-27
remove duplicate infor for Chile avoiding Chile declares war on Chile

Thinking-waffle:


Greatexperiment:
Update the notes as per usual.

PR #687, opened 2019-07-06
fixing bug in bulgaria

Thinking-waffle:
yeah you forgot a bug


Greatexperiment:
This bug has been pulled.

PR #712, opened 2019-07-20
Changed Laos and Cambodia's starting ideology and popularities

BlueFoxyLeMage:


Greatexperiment:
This seems fine, but I am disturbed by the lack of butts.

PR #715, opened 2019-07-20
Redid the entirety of Carpathian Ruthenia's GFX plus some others

BlueFoxyLeMage:


Greatexperiment:
When you're less well known than 'Carpathia Ruthenia' and get called 'others' you know you've really messed up as a nation. 💯

PR #717, opened 2019-07-21
Small fix on decision categories

PordyHUN:


Greatexperiment:
Pordy, you never do anything small. Be proud.

PR #720, opened 2019-07-22
Rokim Strikes Back, with States!

Rokim21:
Tibet now owns Xikang
South America have impassable states in the Amazon region
Brazil's Victory Points are now fixed, finally
Some state names changes, nothing big


history/states/309-Dutch SA.txt
```diff
@@ -1,31 +1,31 @@
...
-
-	provinces={
-		1928 10906 12828 
-	}
-}
+
+state={
+	id=309
+	name="STATE_309"
+	manpower = 15110
```
PordyHUN: It seems you missed a number. instead of 150k the new mapower really 15k?

PordyHUN:
Without knowing the situation of brazil, i would like to ask, why that so many new 1 province state?
It could be done with 1 bigger one?

history/states/309-Dutch SA.txt
Rokim21: I might have commited a little mistake here. I had to give the new states some manpower, maybe I did a mistake along the way.

Rokim21:
I did so to keep the original shape of the states, the impassable states are supposed to be taken together with another state:
Amazonas with Tefé and Santa Inês-Bacabal with Maranhão.
Initially I though of making that province part of the a big impassable state for Brazil, but it would look horrendous if anyone didn't take both Maranhão and Amazonas.

PordyHUN (approved):

Greatexperiment:
I am not a fan of impassable states at all, and would like this to only be a temporary solution. Please look into other methods of limiting the use of provinces in the jungle -

PR #815, opened 2019-08-27
Game rules cleanup.

eliasolsson:
Getting rid of some things that don't work. Simplified focus tree selection a bit for the future. Fixed and tweaked Italian historical AI. Fixed some errors in the log.


Thinking-waffle (approved):
Welcome back into active service private!

PR #994, opened 2019-10-25
Added a Portrait for Tannu Tuva Communist Leader

Rokim21:


Thinking-waffle (commented):
Found that it was had wrong file path, changed LIB to TAN

Thinking-waffle (approved):

PR #1010, opened 2019-10-30
North American Impassable States

Rokim21:


Thinking-waffle:
the extra state should probably be replaced by custom impassable area tha would be easier to update in the future.

Thinking-waffle (changes requested):
Change the extra state into custom impassable area like what I did in central asia it takes a bit of time but is much easier to update in the long run.

SpicyAlfredo:
I need to see the map changes. As there was rarely a tundra terrain that the combined army core of engineers of US and Canada couldn't overcome.

PR #1013, opened 2019-11-01
Many fixes for China and Japan

BlueFoxyLeMage:
- Reorganized the democratic tree for China, giving it more bonuses
- Changed the effects of spirits and focuses for China
- Fix for the "Restart the Civil War" focus for the PRC
- "Government Reshuffle" and "Boost the Moral" (PRC) are now mutually exclusive
- Fixed Rhee's portrait for Korea
- Fixed the Imperial Federation focus being unavailable because UK is democratic and South Africa is neutral
- Japan can't attack China if they're in a faction with them
- New focus icons for Japan and China
- Fixed small saving problems on Lemon's Hirohito and James Gavin


Thinking-waffle (changes requested):
Why did you remove the Greek decision to rename northern Epirus?

PR #1016, opened 2019-11-03
Added opinion modifier for the Soviet Japanese Sakhalin oil agreement

Thinking-waffle:


Thinking-waffle (commented):
It's an AI fix and most content are actually copied from USA man the gun strategy file to make them more reactive and use the ship designer.

Greatexperiment:
You're on thin ice, buster! 

...but if it's new content for a fix - sure.

PR #1046, opened 2019-11-13
Updated descriptor.mod

eliasolsson:
After creating mods with the launcher and studying how it does with the .mod files, this is what I got. Still greatly confuses me though.


SpicyAlfredo (approved):
Yeet

PR #1096, opened 2019-12-30
So many things you will not believe it.

Thinking-waffle:


SpicyAlfredo:
Attack ships on fire off the shoulder of Orion. I watched C-beams glitter in the dark near the Tannhäuser Gate. All those moments will be lost in time, like tears in rain. Time to die.

eliasolsson (approved):

PR #1154, opened 2020-03-19
LAR Integration Part II

PordyHUN:
I Fixed the errors i caused in the decisions, and progressed further.


Thinking-waffle:
History modifications are outdated (when related to leaders not techs or ship variants), reproduce paradox mistakes or some dubious choices that we don't want to make.

Thinking-waffle (approved):

PR #1159, opened 2020-03-27
1.9.1. Beta Patch 3 changes

PordyHUN:
Beta patch 3 come out in March 25.
I applied those changes on our vanilla overwrite files.
Nothing serious.

Removed some vanilla localization file, and moved the extra localization content to our own r56_ prefixed files. We can avoid that in the future something changed in vanilla we miss that.


SpicyAlfredo:
Looks good, Brownie points if you get the game rule switch to work with the history file

SpicyAlfredo (approved):

PR #1241, opened 2020-05-08
Neutral Boosting

Merterbod:
Allow boosting for neutral ideologies (similar to the vanilla 1.9.1 update). Especially with addition of all the alternative monarchistic options for majors, neutral is not longer just "neutral" but a faction for its own.


SpicyAlfredo:
Will add this ourselves this patch

PR #1276, opened 2020-06-05
Plutozz returns!

eliasolsson:
* Removed "Change Scenario" button and a few other things related to 1939, since it doesn't exist anymore.
* Added alt-history German AI behavior paths and made them all reactive. (they adapt depending if you have the standard or 56 tree selected)
* Added some missing form Benelux MTG events. (they still need work to adapt if VLA/WLL is around but the important thing is fixed)
* Minor loc tweaks.


SpicyAlfredo (approved):
Welcome back

PR #1320, opened 2020-06-25
Spicy content Chile first beat

SpicyAlfredo:
Debug isn't cleared and AI needs to be taught how to play new Chile tree but I need to get this and mod on a fresher main build to further development of Chile and start work on AI


common/decisions/r56_CHL_decisions.txt
```diff
@@ -0,0 +1,46 @@
...
+		is_good = no
+
+		timeout_effect = {
+		if = {
+			limit = {
+				has_country_flag = r56_CHL_minor_uprising
+			}
+			country_event = chile.39
+		}
+		if = {
```
featherus: You should use else here instead or another if for the alternative.

featherus (approved):

PR #1392, opened 2020-08-24
Tooltip Cleanup with new feature added in 1.9

SpicyAlfredo:


Thinking-waffle (commented):
Irish docking right should probably be put behind dlc

Thinking-waffle (approved):

PR #1420, opened 2020-09-20
Update CHL_l_english.yml

Arcangelus:
Some extra typos and spelling.
Added missing CHL_supress_the_nacistas_desc. It's no quite historically correct as the party never got widespread support and was mostly dead by 1939 (officially died in 1941), but it gets the point across.
Rewritten CHL_reestablish_the_republican_guard_desc because honestly I doubt many people have any idea what the Republican Guard even was, much less what reestablishing it means. 
Please, someone check those last 2. It seems fine to me, but I'm not a native speaker.


SpicyAlfredo (approved):
good job

PR #1451, opened 2020-10-13
Rename corsica.txt to Corsica.txt

DanielNoga:
Fix case for case sensitive Linux, it should fix different checksum between Windows and Linux players


SpicyAlfredo (approved):

SpicyAlfredo:
thank you

PR #1460, opened 2020-10-20
ideologies flavor

SpicyAlfredo:


Thinking-waffle (changes requested):
oligarchic_democracy instead of oligarchic_democarcy

Could also be used for LIberia and its weird one party rule.

PR #1468, opened 2020-10-27
Spr improvments

SpicyAlfredo:


Thinking-waffle (changes requested):
Couldn't you use base game republican/nationalist cosmetic tags so the unique models are used by dlc owners

SpicyAlfredo:
> Couldn't you use base game republican/nationalist cosmetic tags so the unique models are used by dlc owners

You can but getting it to work is a can of worms I don't feel like opening and getting to work for the three people who use DLC and use our tree. What's committed is the best I'm going to do

PR #1569, opened 2021-01-30
Unique Aviation focuses for Lithuania

lolman-warrior:
I've added some aviation oriented focuses and ideas (if you don't like my ideas, somebulgarian ones will do just as well). I've accidently didn't upload the actuall national focus file itself but afaik it should be fine now


SpicyAlfredo (approved):
Thanks for the work

PR #1621, opened 2021-02-16
Shorter focus duration for Turkey and Netherlands

lolman-warrior:


SpicyAlfredo (changes requested):
Reduce too many

lolman-warrior:
is it centrist enough now?

PR #1624, opened 2021-02-17
Waffle february 2

Thinking-waffle:
Fix French Canada puppet having no name.
Fix bad negation making R56 Turkish tree not visible


Shizophrener (approved):
Seems like your turkey fix didnt go through

PR #1655, opened 2021-03-16
Shifting HON Industry Focuses to the right

lolman-warrior:
I am making way for some monarchist scw focuses


SpicyAlfredo (commented):
slide to the right

PR #1672, opened 2021-03-25
SCW Honduran/New-Spain Focuses and Events

lolman-warrior:
I've added some SCW focuses for Honduras to support the Carlists. I've restricted those focuses only for the scenarios when mainland Spain has the lar focuses since there is already plenty of content when it comes to Spain-NewSpain interactions in the rt56 focus tree, apart from the focuses themselves I've added a couple of events and some ideas


SpicyAlfredo (approved):
Use allow_branch = { } in the futrue for DLC conditions. Just cleaner

PR #1676, opened 2021-03-26
Iraq localisation and few changes

ilkober:
Localisation for Hashemit focuses as well as for a few missing ones
Buffed air branch to match generic one; added few research buffs here and there
Fixed a bug present in event files


Thinking-waffle (changes requested):
add the e to Hashemite, plural Hashemites

Thinking-waffle (approved):

PR #1677, opened 2021-03-26
Minior Fix

lolman-warrior:


Thinking-waffle (changes requested):
Isn't 365 days too short for volunteers?

lolman-warrior:
Those volunteers are intended to represent Carlist emigres from scw, I could change them for a longer period but a belive that it would be too strong of a buff for honduras

Thinking-waffle (approved):

PR #1693, opened 2021-03-31
Lithuania Mission Fix

lolman-warrior:


Thinking-waffle (changes requested):
How is LIT_cadets_return_from_abroad_mission supposed to work if it's available when LIT doesn't exist and at the same time requires you to be Lithuania so which is it?

lolman-warrior:
It a mission, available "completes" the mission, and since the rewards are granted to the player through timeout of mission, completing the mission would equivalent to a fail, hence the change

lolman-warrior:
I am sorry for the weird explanation but it works perfectly with the adjustment, trust me, I've tested it and modeled after "Save Alcazar" Spanish Mission

Thinking-waffle:
I forgot how missions work. And honestly, they work in a very strange way.

Thinking-waffle (approved):

PR #1694, opened 2021-03-31
Fantom march 6

ilkober:
- Albania diplomacy requirments adjusted; few national spirits nerfed
- Iraq focus tree second time readjustment - now anti-colonialism belongs to the politics instead of an army
- Also few bugs/faction betrayals fixed


Thinking-waffle (changes requested):
are you sure that +0.03 is enough to make them flip? Especially as Albania's time is counted.

Thinking-waffle (approved):

PR #1715, opened 2021-04-08
Descs for SAF Commie Path

ghost:


LemuelSora (changes requested):
PR causes many things to have duplicates.
Example: PER_prime_minister_forounghi_desc:0 would exist twice.

PR #1728, opened 2021-04-27
Update INS - Indonesia.txt

gansbeud:
I suggest this edit because I see this file was slightly messed up


gansbeud:
I need to review the changes

PR #1756, opened 2021-05-12
Loc fixes to HOL and enhanced London conference behaviour

marijn211:


marijn2112:
TBA: New event for France and 2 new event options for France and Britain to guarantee Yugoslavia instead of forcing them into a faction, no opinion modifier if Yugoslavia is fascist

marijn211:
Closed for redo

PR #1763, opened 2021-05-23
Communist YUG dipomatic restructure+minor SOV fixes

marijn211:
Creates a bigger split between going with the Soviets vs own faction and splits off the joining of a different communist faction in a new separate focus that filters for European nations
This also moves Soviet economic support down and changes it slightly, so it is less appealing to get the Yugoslavian communism focus but then still join the Soviets anyway
image
(image is without having the string file loaded)

Also includes an unrelated simplification (also removing a bug) of the soviet smersh focus availability check and stops a non-communist SOV (most often as puppet) from boosting communism with Baltic Security


lolman-warrior (approved):
I approve and see no issues with the pull

Thinking-waffle (approved):

PR #1784, opened 2021-06-03
Bugfixes+minor tweaks+altered 2nd Ljublana award

marijn211:
Italy will only take the Aegean Islands and Crete instead of taking basically all of mainland Greece if Bulgaria is involved in the 2nd ljubljana award, Italy's event fired only rarely because it needs to do a focus that already gives them a wargoal on Greece, I swapped some requirements to make this focus relevant for now as Italy still has only one alternate branch, this is a-historical but -before it happened rarely that Italy got anything here -it was too OP if they did -it was already ahistorical kinda because an immediate war choice would be more accurate I guess
‎
‎
Also puts a state check on Greece's R56 Dodecanese demanding, they cant demand it if they gave their islands to Italy

Also makes a focus in Trotsky's tree boost communism so you aren't stuck on down to 50% support depending on what event decisions you make during the coup (0.10, would seem too high but this is scaled down to 0.02)

Makes Albania less likely to give in to being puppeted by Greece, the last factor comparison was up to 2000:30, now its closer to 600:30 at most and down to 120:60 at least when Albania is at low strength

Locks Britain from able to do Scandinavian intervention right from the start because of R56's different ideology values for Sweden

Allows communist Yugoslavia to cancel (keeps progress ig) coup preparations on a country if that country joins their faction (due to war most likely)(not done by AI, unknown if thats an issue in reality), also shows Turkey can be influenced right away when doing the 2nd balkan influencing focus, but still locks it behind doing the 3 required focuses for the diplomacy focus.

Related optional changes that could later be made: tweak Albania's AI thing with Greece puppeting it even more in favor of independence with more/different factors, the strength factors will almost always be in favor of Greece, reducing variety


Thinking-waffle (changes requested):
Don't accept this until after the monthly update, requires testing and marijn proper inclusion into the team.

marijn211:
Should only need 1 additional change namely changing the vanilla/DLC tree to also use my altered 2nd ljbuljana award

ilkober (commented):
Unsure about the changes to events/Germany.txt and "new" Greek event
The rest looks fine and has Fantoms seal of approval

Thinking-waffle (approved):

PR #1812, opened 2021-07-01
Couple of fixes

marijn211:
Also is a non-generic Game Rule description for commi Yugoslavia

Small edit, but am gonna be doing other stuff for halve a week so is maybe usable for the next update whenever that is released


Thinking-waffle (approved):
spot-on, keep it up

PR #1835, opened 2021-07-26
Minor enhancements

marijn211:
-Hide ceding territory that you don't yet own (reduced clutter, also hides when Slovenia is still needed for a focus)
-Fixed demand Slovenia bypassing if Yugoslavia makes it a puppet and add WLL/VLA wargoals _for non-R56 tree_, `will_lead_to_war`'s for WLL and VLA too in both trees
-Conditions for China to request Allied help through their focus tree (_its event I still need to change, but, given how annoying the intervention can be, it should still be better than previously regardless_)
-Another minor Turkish Ottoman restructure (I moved one in text too, might look weird in comparison)
-Germany can no longer do War with Greece when they are already at war with them _for non-R56 tree_ (wrong brackets)
-Minor inconsistency between title and description of the mining continuous focus


Thinking-waffle (changes requested):
This is a fork and could cause problem on integration.
A clean branch must be made out of deciiousmods:Beta-Edits and reintegrate those changes.
It means that the changed files can be saved preventively and then integrated into the new branch ensuring a smooth merge.

marijn211:
Moved to branch

PR #1860, opened 2021-08-15
Added a victory point for Estonia

Deadbirdo293:


Thinking-waffle:
Probably don't accept this PR for now

PR #1866, opened 2021-08-24
September-2-adventure

ilkober:
Fixes for VEN so the tree is not broken


TheBrit7:
Missing localisation for socialist education focus.

Thinking-waffle (approved):

PR #1869, opened 2021-08-26
Improves France's ability to research for heavy tanks

ma44:
Makes the focus that gives a heavy tank research bonus a bit more power into it; a 100% research boost seems very "eh" considering that not only is there a +400% research penalty per year ahead of time for 1941 heavy tanks, but that "historically" France emphasized on distributed heavy tanks and the Germans being able to research heavy tanks faster than France seems very eh.

The French's medium tank research bonus for the 1939 edition of medium tanks probably needs a look over considering it's a significantly more powerful research bonus compared to the German's ability to do medium tanks.


TheBrit7:
France has already been buffed a lot particularly in their tank design for the next patch and I seriously don't recommend buffing them anymore, the focus is perfectly fine as it is right now. And well, even if the focus did need a buff, the changes you made are ridiculously overpowered and would completely throw things out of proportion, I simply don't recommend this PR should be accepted.

ma44:
Reframed it to be more of a reasonable bonus when comparing to other research bonuses;

The reason why this PR mainly exists is to allow France to have an option of going heavy tanks; the current research ahead of time penalties is absolutely massive and practically makes France unable to do 1941 heavies until around mid 1940 even with the 100% research bonus, which is far away enough for France to usually become capitulated.

Readjusted research bonus is to allow France to at least be able to get the 1941 heavy tank technology by mid 1939; giving it a bit of room for producing tanks, while this is still pretty behind compared to German tank technology, it's still early enough to maybe get some tanks out.

AbbusF (approved):

PR #1870, opened 2021-08-26
Soviet defense of the motherland tree and other minor changes

TheBrit7:
Added in the defense of the motherland path i made, and tried to buff the Trotsky path a little.


Darkneon312 (approved):
your mom

PR #1888, opened 2021-09-01
Waffle september 4r

Thinking-waffle:


ilkober:
This PR has Fantom's seal of approval.
Few localizations to ideas (LIT) can be added, but this can be left for another day.

PR #1889, opened 2021-09-02
New states middle east

TokeyWartooth:
Added 4 States to the Arabian Peninsula
Buildings - Supply Regions Adjusted
Unitstack positions slightly adjusted.

Added 1 island state in Socotra (previously part of Aden - South of Yemen)

Split Deir-az-Zur into two states.
The idea came because Kurdistan's historic claim only extends into the North part of the Region.

Made the "Pastoral" State Category Slightly Darker to increase clarity when viewing the States Overlay.
Now they shouldn't appear the same colour as "Town".

Edited Formable Nations Decisions for Form Arabia and Ottoman Conquest in Arabia


SpicyAlfredo:
Need to communicate with Feathers for new states

SpicyAlfredo:
I'm going to close this request for now. Save the branch and work with feathers for the new states post DLC

PR #1905, opened 2021-09-20
Fantom october patch 1

ilkober:
Many little tweaks encountered and shown by VS Code as errors/possible issues


Thinking-waffle (approved):
Impressive

PR #1912, opened 2021-09-27
Simpler typos

tobiasjcat:
Mostly doubled words i.e. "of of" because they were easy to grep for


marijn211:
Nicely done, I have checked it all I think and noticed one mistake and two things that aren't wrong but I think the original was better:
In  localisation/mtg_events_l_english , I think "set in in the business community" was better, you won't wrongly read the "in" as part of "set in", and set seems a less appropriate verb in the context.
In localisation/HOL_l_english , because only a part of the production is referred to, the double "that" makes sense.
In localisation/ENF_l_english , the only thing that's wrong, is this: "if we fail to combat it may also lead to our downfall", the original double "it" was required for this sentence to make sense.

Of course if you have a different idea of how to fix these than the original that'd also be cool.

tobiasjcat:
Makes sense, thanks for the attention to detail. Reverted those edits

Thinking-waffle (approved):

PR #1932, opened 2021-10-12
fixes and new stuff

TheBrit7:


marijn211 (approved):
Looks good, slightly off recommended skillpoints/level but not by far and common for China alr

PR #1944, opened 2021-10-26
November-2-little-changes

ilkober:


SpicyAlfredo (changes requested):
Don't remove blocked-out place holders, they cost literal bits on the file size and there are better places to do cuts for KB's like putting together some common files. Don't cut stuff that's vanilla that's not hurting/helping! Rest looks good!

Thinking-waffle:
is it good now?
Seems to be.

Thinking-waffle:
let's go

PR #1947, opened 2021-10-28
Fix lots of support to anti communists and antifascists

Thinking-waffle:


common/on_actions/r56_on_actions.txt
```diff
@@ -3244,6 +3244,65 @@ on_actions = {
 					remove_ideas_with_trait = staunch_monarchist
 				}								
 			}
```
Arcangelus, line 3244: This is just a suggestion, but you can condense both actions if you do this:

			if = {
				limit = {
					ROOT = { original_tag = CAN }
					FROM = { tag = ENG }
					ENG = { has_government = neutrality }
				}
				set_party_name = {
					ideology = neutrality
					name = "Imperial Loyalists"
					long_name = "Imperial Loyalists"
				}
				if = {
					limit = {
						OR = {
							date > 1940.2.11
							has_country_flag = r56_can_john_buchan_death
						}
						NOT = {
							any_other_country = {
								has_country_leader = {
									name = "Alexander Cambridge"
									ruling_only = no
								}
							}
						}
					}
					create_country_leader = {
						name = "Alexander Cambridge"
						desc = "POLITICS_JS_WOODSWORTH_DESC"
						picture = "gfx/leaders/CAN/r56_portrait_CAN_Alexander_Cambridge.dds"
						expire = "1965.1.1"
						ideology = oligarchism
						traits = {
							governor_general
						}
					}
				}
			}
`
For this suggestion, I made 2 assumptions:
- The party name is supposed to always change when Canada is made a "neutral" British puppet.
- The state of "John Buchan" is only relevant to know if "Alexander Cambridge" should take over or not.

PR #1948, opened 2021-10-29
Update raj_on_actions.txt

Arcangelus:
Corrected a scope to what I believe was intended. What was here before was ignored, as it was checking the tag of the province itself rather than from any of the 2 nations involved. It also gave an error, that happens to be the reason I noticed in the first place.

### DIFF
--- a/common/on_actions/raj_on_actions.txt
+++ b/common/on_actions/raj_on_actions.txt
@@ -199,7 +199,7 @@ on_actions = {
 				FROM.FROM = {
 					if = {
 						limit = {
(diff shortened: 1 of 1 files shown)


Thinking-waffle (changes requested):
This is literally doing nothing

Arcangelus:
After trying a few times, I'm unable to reproduce the original error.
As such, I'll close this.

If I happen to encounter it again, I'll find a way to reproduce it before posting anything else. Sorry for the trouble.

PR #1986, opened 2021-11-26
[DO NOT MERGE] Find + Replace's old numbers associated with different states into their new numbers

ma44:
Looked over a little bit but considering there's inaccuracies like West Virginia becoming a different number than expected, this will definitely require some testing.

States like Ostrhineland being changed from 807 to 848 requires a passover the focuses and events that utilize such old numbers and replaced with new ones, which this PR does.


ma44:
Will need to resolve conflicts and recheck the state IDs once more with Waffle's list

marijn211:
in "POL_Border_Conflicts_Gdansk.txt" it looks like you switched up Gdynia and tiny Danzig, same target and source state, this mixup also appears in other files like Germany's focus tree under demand Memel
The changed number in "common/defines/r56_defines.lua" is not a state and should be reverted

common/defines/r56_defines.lua
```diff
@@ -54,7 +54,7 @@ NDefines.NTechnology.BASE_YEAR_AHEAD_PENALTY_FACTOR = 4.4 -- Base year ahead pen
 NDefines.NTechnology.BASE_RESEARCH_POINTS_SAVED = 56.0 --#base game 30 changed to 50 and then why not 56?
 
 -- Air Combat
-NDefines.NAir.AIR_WING_XP_LEVELS = {20, 40, 60, 80, 100, 140, 180, 220, 260, 300, 390, 480, 570, 660, 750, 780, 810, 840, 870, 900}
+NDefines.NAir.AIR_WING_XP_LEVELS = {20, 40, 60, 80, 100, 140, 180, 220, 260, 300, 390, 480, 570, 660, 750, 780, 85, 840, 870, 900}
```
marijn211, line 57: Pog, test comment, this line should be reverted

ma44:
Already covered by Fantom just a couple of hours ago

PR #2001, opened 2021-12-16
New VP's and Important lesser USA Railways added. Alaska and Erie RR main lines 

SpicyAlfredo:
Erie and Alaska Main Lines


marijn211:
Only victory point name with 2 cities? Seems a bit inconsistent, I'd image there are more candidates then

SpicyAlfredo:
> Only victory point name with 2 cities? Seems a bit inconsistent, I'd image there are more candidates then

Yes, the area is generally called that in reference to both cities at times. The tile includes the two towns so it made sense to use that naming scheme

marijn211 (approved):
Playtested and read through, didnt seem to have errors and everything showed up

PR #2005, opened 2021-12-21
Fixes for NSB (accept when many)

ilkober:


SpicyAlfredo:
accepting so this isn't a stale pull request

SpicyAlfredo (approved):

PR #2007, opened 2021-12-24
SEA and Xinjiang/Tunganistan loc changes

RMD-Angel:
Added namelists to Singapore and Malaysia.

Added a monarchist option for The First Sila and one cosmetic tag for each except the historical one. Note that each cosmetic tag is named after the neutral party, and oppositions of other ideologies would naturally be of opposite nature (religious opposition for a secular ruling party and vice versa).

Renamed fascist and neutral Tunganistan into it's English translation for consistency.

Added the option of going to the East Turkestan Republic under the Niyaz presidency or Kumul Khanate under Bashir as democratic and fascist Xinjiang respectively for variety. The Chagatai flag is used for the latter because there is no flag of the Khanate and they are descendants of the Chagatai. In turn, Tunganistan now has Elihan Tore as a democratic leader.


TheBrit7:
Please fix the conflicts first.

PR #2018, opened 2021-12-31
France character conversion

ma44:
TODO:

- [x] Fix up the GFXs
- [x] Look at "No Step Back" changes and port all the different French character changes from there
- [x] Update focuses to support new character system


marijn211:
You should remove converted ideas, also next time use a branch

TheBrit7:
> You should remove converted ideas

Ideas do not necessarily need to be removed, they can just have their name changed to avoid conflict errors. It is good to keep track of the legacy advisor and also can be used to use the legacy advisor unlock tooltip if needed. But i completely understand your point.


Is this finished?  I don't really know as I don't see any Free French character adjustments, particuarly the ones gained through events, however i may be wrong as i just took a quick look.

TheBrit7:
I'm going to close this until all characters have been properly converted. Open a new one then.

PR #2021, opened 2022-01-02
Egyptian characters 

TheBrit7:


marijn211 (approved):
Everyone showed up with image and no error log messages other than from the placeholder I commented out

PR #2059, opened 2022-01-21
GFX tweaks, should lower the amount of same ideas + little bugfix

ilkober:


ilkober:
This is the last PR from this branch, I swear

Thinking-waffle (approved):

PR #2067, opened 2022-01-29
France character conversion 2.0 but with vichy/free french

ma44:
Added in a new commit to update vichy/free french events into the latest NSB code; it still appears to work with just the name being a string rather than a token when called with console commands so it should be alright


TheBrit7 (approved):

TheBrit7:
Seems alright, but will need some testing.

PR #2093, opened 2022-02-19
Mexican characters and white Russians

TheBrit7:


TheBrit7:
I forgot, but this also includes Colombian and a bit of Commie China's too.

Thinking-waffle (approved):

PR #2098, opened 2022-02-22
Minor Fix

LoLNN2:
a few flags switched to their correct versions


marijn211:
Already fixed these on Waffle's branch

LoLNN2:
True

PR #2128, opened 2022-03-09
Changing the ROC color

TheCrazyAnarchist:
ROC becoming blue to avoid the subtle and perhaps accidental racism of China being yellow.


ilkober:
Don't France and French Indochina have similar colour + Xibei San Ma is also blue, so one could end up with three/four blue nations bordering each other lalmost from the start of the game?

TheCrazyAnarchist:
This should be a unique enough blue, most similar to Sinkiang.

PR #2152, opened 2022-03-20
Big batch of loc fixes.

PetrogradXXII:


marijn211 (dismissed):
Don't see any other mistakes, but there's a lot of unnecessary changes imo
For example replacing adjectives like "Swedish" with dynamic ones, it's already very clear who is being referred to, no way the dynamic is gonna be clearer unless appropriate in political context. The punctuation or word order changes might be more correct but it doesn't become any more readable.

SpicyAlfredo (approved):

PR #2153, opened 2022-03-24
Filling in the gaps with generics

TheBrit7:


marijn211 (dismissed):
The first commit "y" seems to revert recent changes to the main branch, maybe this commit could be reverted if there's nothing else there

marijn211 (approved):
Code looks fine, tested and any error log seems unrelated

PR #2157, opened 2022-03-26
Pepper peru

pepper419:
i just nerfed the national spirits of Peru and increased days to the foci that i think are not necessary to last so long, the "rare" part of the tree is necessary and that they are alliance options for the country and the focus that are without an icon It is an error that I do not know since it supposedly grabs generic icons, i added the events


ilkober (changes requested):
Localization needs to be rewritten (Capitalization)
Decisions are either missing or their tooltips are not defined
Some icons are missing
Would be nice to havre focus tree layout adjusted
Bugfixing needed + nerf of some OP focuses.

TREE IS NOT READY FOR MERGE

PR #2158, opened 2022-03-27
Pepper peru

pepper419:
i just improved the locs and added new icons for Peruvian national spirits and a new focus for the communists, i wanted to move the democrats so that the final part is not so messy but I couldnt since it messes up the tree in an incredible way and i think it would be better leave it like that, add the loc of the unlockable descriptions when unlocking a focus


marijn211:
Needs fixing, please put as draft pull request

SpicyAlfredo:
Im going to close this pull until it done

PR #2206, opened 2022-04-11
ADR paths + character fixes

eliasolsson:


TheBrit7 (approved):
Andorran AI paths are something we never asked for but always wanted, true hero.

PR #2230, opened 2022-05-05
I've discussed with admin

Khorsun:
I talked to @Thinking-waffle, he should be know about of this. We have details settled.


ilkober:
I think you can remove the file "Na tot samyj sjuchaj" - it seems not to be needed

Khorsun:
sorry fixed

marijn211:
Instances might be broken, since global flags have a tendency not to work in allowed's because they only load on game start

PR #2253, opened 2022-05-22
Fantom june 2: fixes and fixes and beyond

ilkober:


marijn211 (approved):
Small error log decrease, files look fine, French tree works again

Issue #2251, opened 2022-05-19
CV Jet Fighters only have Air Superiority/ Interception missions


kmjohnsonpgh:
As you can see below, CV Jet Fighters are shown to have the same missions as regular fighters/ jet fighters on the research page. However, when they are used to create an air wing, they are only able to perform the air superiority or interception mission. I believe they did have the correct missions available in the past at some point, so I figure this is a bug as opposed to an intended design change.

![Research Page](https://user-images.githubusercontent.com/105891846/169416119-9d29b9dc-4193-4d39-9013-328d41738112.jpg)
![Air Window](https://user-images.githubusercontent.com/105891846/169416120-01c0bef5-deae-4e18-9c13-6d096ca276cc.jpg)

AbbusF:
![圖片](https://user-images.githubusercontent.com/50837019/171259835-aed5600d-4b95-47d9-b1ce-8671f923c171.png)
Thank you for your report, it will be fixed in the next patch.

PR #2270, opened 2022-05-31
Issue 2251 cv jet fighter missions

tobiasjcat:
### Summary
Issue: #2251  
[kmjohnsonpgh](https://github.com/kmjohnsonpgh) noticed despite having values for ground and naval attack, cv jet fighters were not able to perform the missions which r56 fighters can.  

Taking a look, I noticed that all r56 cv fighters have values for ground and naval attack but regular cv fighters were **also** not about to perform these missions.  
This looks like it is caused by cv_fighter units not having the `cas` type. I added the `cas` type to `cv_fighter` in air.txt. I'm not sure why `cv_jet_fighter` did have `cas` but `cv_fighter` did not. I assumed it was supposed to since `cv_fighter_equipment*` has cas values.

I also removed the `cv_jet_fighter` unit since it appears to not be used. I assume this is because `jet_fighter` uses a different equipment archetype as `fighter` so that fighters don't upgrade to jet fighters, which is different behavior than CV jet fighters which do replace prop cv fighters.

### Testing
I tested this locally on my own machine with no other mods than this branch of 1956_beta and it doesn't crash. Here's what that looks like  
cv_fighters

I did not test these changes in actual combat or naval combat, which someone should do eventually.

### DIFF
--- a/common/units/air.txt
+++ b/common/units/air.txt
@@ -35,15 +35,6 @@ sub_units = {
 		need = { jet_fighter_equipment = 1 }
 		categories = { category_fighter }
 	}
-	
-	cv_jet_fighter = {
-		sprite = jet_plane
-		priority = 1
-		active = yes
-		type = { fighter cas }
-		need = { cv_fighter_equipment = 1 }
-		categories = { category_fighter }
-	}
 
 	cas = {
 		sprite = light_plane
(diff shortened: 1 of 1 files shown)


SpicyAlfredo (approved):

SpicyAlfredo:
Thanks for the input

PR #2277, opened 2022-06-06
3-way civil war for Germany and bugfixes

marijn211:
* Going communist or democratic leads to a highly customized civil war (and disables placeholder Rhineland decision for them)
* Democratic focus to attack communist Benelux/France, allying Czechoslovakia focus, alliance instead of annexation of Austria option which also allows for allying Italy or teaming up with fascist Italy against communist Yugoslavia.
* Going fascist Germany now hides the monarchist block
* Unrelated bugfixes


ItsNotPizza (approved):
Communist Side is  a little hard, but other then that looks great

PR #2288, opened 2022-06-18
Abbus' June Not-so GFX Adventure 2

AbbusF:


marijn211:
Seems to have no bugs, but Garrison might be a bit too strong except the movement penalties which might be a bit too much

marijn211 (approved):
Discussed the nerfing of garrison, should now be pretty balanced

PR #2289, opened 2022-06-19
Altered values for going bye-bye mussolini

SwedishNorwegian:
Added 35% Auth-check for being able to do Assassinate Mussolini along with increasing value from completing the focus from 10% to 15%


marijn211 (approved):
Discussed in VC, all good

PR #2303, opened 2022-06-27
Abbus' June Adjust Adventure

AbbusF:
-Mot. Shocktroop reduce supplies usage from 0.8 to 0.08
-Nerfed most PRC idea to halve or more


marijn211 (approved):
No more extra error log, cant really test all of this

PR #2308, opened 2022-07-01
Peru for July Beta?

Thinking-waffle:


ilkober:
I thought we usually PR focus trees in the middle of the month so they can be fixed/tweaked with enough time
Seems a bit rushed right now

PR #2312, opened 2022-07-02
Romania rework

TheGameAce:
Ignore the history file changes and the dev tree, everything else should be fine.
*Might also need to activate the current non-DLC tree that I included.


marijn211:
Not in a PR-ready state, industry branch looks nice, but it needs to be well-integrated into the existing tree if the political part isn't done.

PR #2316, opened 2022-07-03
Removed Adv Light Tank from Modern Tank req.

SwedishNorwegian:
This change is one made from PDX and has yet to be implemented in our mod.


marijn211 (approved):

marijn211:
Btw if anyone wondered if we transferred the flame tank nerf, we don't override that file so we don't need to

PR #2318, opened 2022-07-05
Romania rework

TheGameAce:
Everything should be set for the pull this time. No state errors, idea error fixed, etc.

IMPORTANT NOTES:
There are 3 focus tree files right now. The dev version is what will appear on your end when booting up, because it's what I'm actively working on on my end. Just disable this so it doesn't override the other two.

r56-romania.txt is the generic tree with my Industrial branch integrated, and should automatically take effect for those without the DLC. Further commits will have this tabbed out so it doesn't interfere with the development version.

romania.txt is the DLC tree, and simply needs to be re-activated. I keep this deactivated on my end to prevent issues. Simply change the add = 0 to add = 10.

Please ignore the history file changes I made. They'll add national spirits that have no way to be removed currently, and will cause issues. Line 194 is the only change that needs to be made, adding quotation marks to the DLC title.


ilkober:
Unfinished and not full
Disables DLC tree for now

PR #2319, opened 2022-07-08
Brit august 1

TheBrit7:


marijn211 (approved):
No error log from renamed focus id's, changed layouts and icons show up fine, loc all looks correct

PR #2330, opened 2022-07-17
R56 Germany stuff and fixes

marijn211:
- R56 Germany historical AI extended from until the invasion of Poland to until operation Barbarossa
- Nerfed a tank idea and fixed a typo in the Road to 56 tree
- AI-only idea so Iceland doesn't join the Axis (or any faction) when attacked by democratic UK
- [DoD] Croatian insurrection has Yugoslavia declare war on Croatia instead of reverse, this stops Yugoslavia's guarantors from joining a war where Yugoslavia actively joined instead of being declared war on. 
- Polishing R56 Germany in general so its ready to be set to default
- Rail map is normal again


TheBrit7:
If you are removing the vanilla tree as the default one and having them both as an option, then which one is the default one?

TheBrit7 (approved):

PR #2342, opened 2022-08-06
Venezuelan AI guidelines and localisation robustion

TheBrit7:
At last, the alternate history Venezuelan AI will no longer be inactive and confused and now actually get down to what their actual objective is.


marijn211 (changes requested):
Reverts recent PRs by broken updating from main somehow

PR #2347, opened 2022-08-07
Added random_owned_controlled state to focuses to not build in them i…

SwedishNorwegian:
Added random_owned_controlled state to focuses to not build in them if Sweden does not own them


ilkober (approved):
Gut

PR #2355, opened 2022-08-19
Old Soviet Tree Gamerule Option

marijn211:
Currently slightly altered so I could easily make it into a submod if you don't like.
Done:
Game-rule, Stalinist purges converted to character system, new foreign policy ported into communist part old tree, most design companies adjusted, events re-activated and made tree specific, advisor checks updated, some guys now use the new exile flag.
To-do:
Check starting buffs and debuffs, expand alt-history politics and expand their foreign policy, some focuses might need unessential character system changes.


marijn211:
Should now be developed far enough for testing, the only thing I know I have to do is change starting ideas, other things could arise from testing, or would be more optional improvements

TheBrit7:
Needs balancing with the current vanilla tree to avoid it simply being used as a meta tree.

PR #2358, opened 2022-08-26
Fixes and Improvements 

Shizophrener:
GUI now operating on controllers instead of owners; some state transfers not updating fixed; regular updates for the warlords


marijn211 (approved):
Did a quick unification run and no error log or bugs

PR #2362, opened 2022-08-29
Pepper peru

pepper419:
I finished putting the description to everything and most of the focus are already balanced in terms of time and effects


ilkober:
Not ready

PR #2371, opened 2022-09-08
Portraits

Enrx3125:


TheBrit7 (changes requested):
Trying to PR commits from 9 months ago without updating them is not a good idea. I'm pretty sure we have a good chunk of these added anyway.

TheBrit7 (approved):

PR #2372, opened 2022-09-13
Fixes for Soviets - The Right Opposition

CZThomasCZ:
- [SOV] Fixed "Return Democracy to the Party" when going The Right Opposition - advisors from the Left were no longer exiled, but still imprisoned.
- [SOV] Removed Lavrenty Beriya's (NKVD Head) check for having Nikita Khrushchev as the country leader - breaks paranoia system when Stalin does the third (last possible) NKVD Head purge.
	- a different possibility is to limit the NKVD Head purge to only happen twice
- [SOV/Spain] Fixed "Covert Support for Spanish POUM" not making the POUM rise up instead of the anarchists.
	- changes in events\LAR_NewsEvents.txt and events\LAR_Spain.txt were copied from vanilla


ilkober (approved):
Seems fine

PR #2374, opened 2022-09-18
Waffle october 1r

Thinking-waffle:


marijn211 (approved):
Currently seems free of additional error log and game rules show up fine, too big and spread out to playtest much

PR #2391, opened 2022-10-28
Update quad_engine_airframe.txt

LittleRedPoint:
The transport plane did not work at all. The missions description was missing, so i added that in my game and after that supply missions started working.


marijn211 (approved):
In quick playtest, enabled air supply on gamestart transport planes, so seems all good

PR #2392, opened 2022-10-28
Update countryoccupationview.gfx

LittleRedPoint:
With Blood alone extension there are 13 country occupation icons. Thus mine was offset so i changed it to 13 and all icons were shown correctly. I don't know if it could be issue without BA extension.


marijn211 (approved):
We dont override the mentioned texture, so making this value the same as vanilla makes sense as fix

PR #2405, opened 2022-11-08
Adding a focus to support the Hungarian revolution for US

SanzhKM:


SanzhKM:
I am not sure if you have anything like that, but the focus icon for the Hungarian Revolution can be replaced with the Hungarian Flag with a hole in middle that was used during the uprising. I mean if you have anything that fits the description.

ilkober:
Could be an interesting idea BUT:
1. Hungarian Revolution was not funded by US as far as I know...
2. Focus layout overlaps with tree we currently have.
3. Maybe make it more generic in expansion of NATO into Europe, why Hungary in particular?
4. The focus sparks a civil war, no matter if it is a player of not... What if a player is doing a communist Hungary game and is dominating Europe? Shall the civil war still be able to fire?
On the other hand, what if Hungary has 1 state only, for example?
What if Hungary is a non-democratic puppet of Tsarist Russia? Shall they also receive a pro-democratic civil war?
What if Hungary is communist, but not in USSR faction? Also civil war shall be possible?

PR #2415, opened 2022-11-14
Spanish translation

criticalplus:
Spanish translation


TheBrit7:
This is certainly impressive, however the is a lot of incorrect capitalisation especially on focus names. Obviously I couldn't comment on the Spanish itself but that seems like something that needs to be fleshed out. Good job though.

criticalplus:
Hello, I have corrected the capital letters and added some missing team translation

TheBrit7 (approved):

PR #2417, opened 2022-11-16
Uzbekistan: Characters

SanzhKM:
Wouldn't hurt...

### DIFF
--- a/common/characters/UZB.txt
+++ b/common/characters/UZB.txt
@@ -1,0 +1,823 @@
+characters={
+
+	####    ####   #    #  #    #  #####  #####   #   #      #       ######    ##    #####   ######  #####    ####          
+	#    #  #    #  #    #  ##   #    #    #    #   # #       #       #        #  #   #    #  #       #    #  #      
+	#       #    #  #    #  # #  #    #    #    #    #        #       #####   #    #  #    #  #####   #    #   ####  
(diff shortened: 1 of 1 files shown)


ilkober:
Characters have no assignment to a country
And they don't have names defining their strings
Plus, leaders have no portraits defined (even generic ones)

SanzhKM:
> Characters have no assignment to a country And they don't have names defining their strings Plus, leaders have no portraits defined (even generic ones)

I mean uzbekistan has no advisors, high, command or generals so ya'll could pick out those you wanna keep, sort of like "assigning" them or recruiting them through _recruit_character.

also tf you mean by "names defining their strings"?

ilkober:
> I mean uzbekistan has no advisors, high, command or generals so ya'll could pick out those you wanna keep, sort of like "assigning" them or recruiting them through _recruit_character.

Since there is no info on them provided, they could be randoms which we will have to check twice instead. And since the translation of poitentially cyrillic alphabet doesn't go smoothly, having additional info would be great. Who are they, background etc. Check some of the other advisors we have, for example like I did in Mongolian file.

> also tf you mean by "names defining their strings"?

Advisors have an entry which is aimed to give them specific names like 
SOV_iosof_stalin = {
      name = "Iosif Stalin"
      ...
}
Otherwise the game will likely either display a broken SOV_iosif_stalin string OR will assign the generic name

SanzhKM:
> 

Get ready for round 2

PR #2418, opened 2022-11-16
Uzbekistan: Characters 2, electric boogaloo

SanzhKM:
Round 2, as promised


ilkober:
Thank you!
Do you mind also including the cyrrilic names in their descriptions?

It is sometimes hard to double-check info on them
Example being the Абдусаттор Эшонкулов which took me a few tries to get the name translated correctly

SanzhKM:
> Thank you! Do you mind also including the cyrrilic names in their descriptions?
> 
> It is sometimes hard to double-check info on them Example being the Абдусаттор Эшонкулов which took me a few tries to get the name translated correctly

and the winner...through a technical knockout......is......"ilkober"!!!

PR #2421, opened 2022-11-18
Uzbekistan: Characters...The Trilogy Final

SanzhKM:
I've added cyrillic translation to some characters, those who do not have a translation are those who you could easily find in Wikipedia or that there are no uniquely russian sources for that character.


TheBrit7:
None of the leaders have portraits defined?

SanzhKM:
> None of the leaders have portraits defined?

No, based on my tests, a generic one will be assigned, unless ya'll want to assign a specific portrait to the leaders (hence its empty); but that's not for me to decide

ilkober:
Should be working fine, needs double checks by some other dev to make sure that portraits are shown correctly

marijn211:
Because this is so old Ill transfer your changes onto a branch we can more easily playtest, and to avoid any issues with this being so old

marijn211:
Advisors icons seem broken

PR #2429, opened 2022-11-25
Custom achievements

Thaldoras:
added mod achievement for RT56
added decision for Argentina to rename the Falklands and South Georgia if they own them.


ilkober:
Before further checks, need to see what game rules allow achievements anyway (fragmentation/defragmentation, path selection, focus tree selection etc)

SpicyAlfredo (approved):

PR #2448, opened 2022-12-09
German civil war recoded to fix some issues

marijn211:
Reworked the way the German civil war works, so players don't crash when opening the officer corps and lose all their equipment designs. Almost completed, only needs war between the different sides to be properly started.
The player stays the original tag instead of being switched to a new dynamic one, the basegame has a bug that apparently causes crashes when such player countries access the officer corps, it also will keep equipment designs.


TheBrit7:
You can put up an actual PR once you are finished.

PR #2451, opened 2022-12-15
Waffle bba 12

Thinking-waffle:


marijn211 (approved):
Checked error log, read most files, checked in-game for stuff the error log wouldn't show

PR #2469, opened 2023-01-06
Mengjiang focus tree

TheBrit7:
I would say this tree is 80% done however it desperately needs polishing and to be made robust - right now it's very railroaded and could probably break easily. The coding is quite poor in some areas and doesn't match vanilla. It is also lacking any kind of army and navy tree. I will be finishing these things when I can find the time or motivation to do so but if anyone wants to pitch in and help get this in the mod without leaving this content to go to waste then feel free, treat it as a side project.


TheBrit7 (commented):
No

ilkober:
Lets add it to the list of projects to finish

PR #2472, opened 2023-01-06
Pepper peru 3

pepper419:
new balance and polish


ilkober (dismissed):
Lower stability from effect
Use advisor highlighting tooltips we already have + apply to all advisors unlocked through decisions/focuses

SpicyAlfredo (approved):

SpicyAlfredo (approved):

PR #2488, opened 2023-01-19
German civil war recode attempt 2

marijn211:
Its almost done, but I need support to fix the war declarations.
The only other bug is that some troops arent teleported properly, but Ill hold off on that until I know the wars can be fixed.


TheBrit7:
Open when ready.

PR #2510, opened 2023-02-06
Spicy Feb last bag of tricks

SpicyAlfredo:
Bunch of small stuff again


ilkober (approved):
Looking good man, looking good
That whole +10% conscription though...

PR #2515, opened 2023-02-07
Norrman communist rework

SwedishNorwegian:
Rework to the Swedish Communist Branches.


ilkober:
Leave it for the next patch

PR #2517, opened 2023-02-08
Branch hiding for France

marijn211:
Hides wide unavailable political branches for **France**, no error log and double checked for focuses that aren't always visible


SwedishNorwegian (approved):
okidoki

PR #2526, opened 2023-02-23
update characters from old style to new style

FOREVERGWC:


marijn211:
It'd be convenient if you could remove the unnecessary indentation changes and comment removals etc.

FOREVERGWC:
> It'd be convenient if you could remove the unnecessary indentation changes and comment removals etc.

I'm sorry, because I use the format converted by the js parser. To prevent errors, all comments are deleted by default. If I add those comments back, there will be no problem, right?

marijn211:
> > It'd be convenient if you could remove the unnecessary indentation changes and comment removals etc.
> 
> I'm sorry, because I use the format converted by the js parser. To prevent errors, all comments are deleted by default. If I add those comments back, there will be no problem, right?

Please discard the parser changes if possible, most stuff was already correctly indented and it makes tabs into spaces, its also hard to see if theres any mistakes because we cant easily playtest your fork

PR #2550, opened 2023-03-06
Last Polish and AI adjustments

SpicyAlfredo:
Next pull will be all bug fixes


SpicyAlfredo:
The reason the ASIC letters on the FRI_english file were removed was that it gave my text editor warnings lol. Never seen that before

PR #2561, opened 2023-03-12
Added a victory point to British Honduras

Altitud49:
British Honduras had no victory point which made the decolonized Belize have a capital with the wrong name. That is now fixed


SpicyAlfredo (commented):
Don't like vps with values of '2' changed to 1

SpicyAlfredo (approved):

PR #2564, opened 2023-03-21
Waffle april 1

Thinking-waffle:
Beware of the Catalinas (and mostly their predecessors) in the USA history file.

Vasyl Ukraine should get +1000 opinion with Habsburg Poland and Otto's AH


marijn211 (changes requested):
Plane icon file has a lot of (new) wrong scope errors
I wouldn't know for sure but it looks like all the cosmetic-tag based icons do not work like that

TheBrit7:
This PR has a lot of errors.

marijn211 (approved):
Despite spamming errors, the cosmetic tag specific plane models do seem to work perfectly fine

Thinking-waffle:
But the SPA_director_fighter and poum_fighter things need to be fixed.

PR #2566, opened 2023-03-23
Removed redundant events

calvin-fisher:
The `r56_hungary_coastline.3` event already does what events 4 and 5 do, but better. Event 4 doesn't check to see if Italy is AI before firing, and only has one option: hand over the territory. Event 3 already has a choice, and reasonably grants Hungary a CB if Italy declines.

### DIFF
--- a/events/r56_hungary.txt
+++ b/events/r56_hungary.txt
@@ -116,59 +116,6 @@ country_event = {
 	}
 }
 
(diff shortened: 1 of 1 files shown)


marijn211 (approved):
These events do seem to fire at the same time about the same thing, and not having any options is silly

PR #2568, opened 2023-03-24
Egypt AI improvements and technology fixes

TheBrit7:


SpicyAlfredo (changes requested):
Moving the RK states of the Goverment government to Ukraine was not the change I was okay with. It needs a new decision to expand the General Government to these new states or leave it as is.

SpicyAlfredo:
Everything else looks good

TheBrit7:
No, because otherwise Generalgovernment wont exist by the time they should, which overall makes them quite redundant. I also recall you saying that I was free to change it as I wish so with such a minor change I far from seeing it matter.

TheBrit7:
Your requested changes should be addressed. Please remove your block.

ilkober (changes requested):
MON_states_required_for_greater_mongolia_flag in on_actions 
is no longer needed since formables have been reworked and the flag is obsolete

TheBrit7:
Done

PR #2569, opened 2023-03-25
Abbus' Apirl '23 Adventure 

AbbusF:


common/ai_strategy/GER.txt
AbbusF, line 1: Role Ratio changing too much will break

SwedishNorwegian (approved):

PR #2580, opened 2023-04-02
Trait for two leaders, should improve AI picking personality cult et al.

Thinking-waffle:


SpicyAlfredo (commented):
Does the ai run calculations on the R56 laws before 1939? It's why it was removed in the first place a few months ago

SpicyAlfredo (approved):

PR #2583, opened 2023-04-04
Fixes to Generic Shared Tree/Malay/FRI

Simpyslav:
Atm: Too many bonuses are added (all created variants vary and are based upon the competitors own aircraft
What to replace old Nav Bombers with? Like its so redudant since there's nothing to replace it especially since theyve main-lined the new tech so much.

But if anyone can spot fuckups would be nice as im blind
Also i know the indentation is all over the place i can't figure out how someone would want it - its somewhat easy for me to see and i don't  know how "preffered" it should be


Simpyslav:
Ah so i realized one mistake being the IF in Bomber Competition and then the FIghter Comp

marijn211:
Fantom changed Saudi Arabia so Ill manually put your changes back in, discarded them on purpose to not accidentally cause conflicts

Edit: Saudi Arabia's air content was (almost?) a generic tree copy which Fantom has switched to the actual generic branches, so it no longer exists

marijn211 (dismissed):
Needs fixes

marijn211:
Converted to draft as the Siam work is WIP and the PR doesnt need to be reviewed yet

PR #2591, opened 2023-04-08
Update r56_countries_l_english.yml

Altitud49:
Added some adjectives which might fix a part of the "_adj Federation" bugs
Fixed lowercase Côte d'Ivoire spellings
Removed the Canary adjective "Bengali" (what was it doing there?)
Dodecanese localisation fixed so that none has to see localisation ids in the middle of their campaign


Altitud49:
Might not be the right solution

PR #2594, opened 2023-04-10
Dodecaneso

Altitud49:
Fixed some Côte d'Ivoire spellings and added an Italian name for Dodecanese so that ids just do not pop randomly when hovering over Rhodes.


marijn211:
I think you forgot to rename the Dodecanese for the Italian version

Altitud49:
I also recolored Slovenia and Macedonia

marijn211:
It seems like youre missing a `"` behind your extra loc line for Monaco (it seems this also applies to loc of the preceding states)
I dont see why we would need to change those country colors to the base game ones, but theyre not terrible either

Altitud49:
It seems that the quotation marks were forgotten during the addition of the Microstates

Altitud49:
The Côte d'Ivoire fix is no longer needed as it has already been fixed in the new countries localisation file, speaking of the new countries localisation, I sorted them in alphabetical order so it is easier to find what we need without ctrl+f, for example, adding the missing countries such as Green Ukraine

Altitud49:
Ok so apparently Mauritania has not the same tag in he first one than in the other
I guess this is heavily in the works I didn't add Green Ukraine nor Hyderabad I just classified it alphabetically revert my changes if I shouldn't have.
I'll just wait for the next update to clean this all up

TheBrit7:
Do that whenever you're ready to.

Altitud49:
Ok I was bored just fixed the file, I figure we can now delete the original "r56_countries.yml"

Altitud49:
Was there an error in my code ? Did I do something wrong ? (please just tell me I do not wan't to keep making the same mistakes over and over again)

Altitud49:
Silly me it didn't even get reviewed...

PR #2606, opened 2023-05-02
BBA User Friendly Air Branches

Simpyslav:
Possible FIx for Most Air Branches?
I'll fix the issues with them happily
Germany should get correct bonuses now
and many other nations now recieve correct bonuses instead of getting 1944 Tech in 1938 because focus tree broken ;D


marijn211:
Free of errors, some of the BBA research buffs seem overly specific but I don't know the available categories

Simpyslav:
Wanted to give BBA users something for "Cas Efforts"
While i don't want them to get AT bonuses (wich are the modules for anti-tank cannons)
So i specified giving them Cannons instead - might be a  mistake but could be avered by possibly unlocking the module itself?

I realize this is semi-obsolete because someone decided to change the categories for earlier bonuses to work for bba users however some of them are ridiculously overpowered thats why i specified BBA to change the bonus for players not to get 1944 tech in early 1939

Another thing was Scout Planes (Wich for LAR users worked fine but for Bba users did get nothing)
and the final thing was the add_equipment wich was replaced for BBA users to actually spawn correct things

Thinking-waffle (approved):
nice progress, could take into account some czech events in the future.

PR #2608, opened 2023-05-04
Light Tank Recon Nerf

DutchPlanDerLindeSus:
Made it so Light Tank Recon can no longer have medium armaments on it even with the Turret Shield turret


SwedishNorwegian:
Historically Gun Platforms were used to mount much larger guns on smaller chassis than they were typically capable of mounting. By removing tank_medioum_main_armament from there you're disabling a lot of historical designs.

PR #2615, opened 2023-05-08
Fixes

marijn211:
-Vichy France can use their focus tree
-Hid an additional French character that is unlocked by a disabled focus branch
-Fixed a vanilla event that had no available option for a country.
-Some error log thingies


ilkober:
Why not make Petain a character per default and promote him as Vichy France arises since it is a civil war tag?

SwedishNorwegian (approved):

PR #2618, opened 2023-05-09
Waffle may 3

Thinking-waffle:


history/countries/BUL - Bulgaria.txt
```diff
@@ -143,14 +143,13 @@ if = {
...
-	else = {
-		set_technology = {
-				
-			early_fighter = 1
-		}
+	set_technology = {
+		engines_1 = 1
+	}
+}
+else = {
```
marijn211: The nested form of if/else does not work in history files

marijn211 (approved):
Additional error log is likely only if DLC is not owned

PR #2638, opened 2023-06-08
Laura's one year long Québec's Tree adventure

LauraPoggies:
full rework of Québec's Focus Tree


marijn211:
Bunch of errors so draft for now

LauraPoggies:
Starting on fixing them, all the ideas GFX are currently being fixed

ilkober:
Note: if error log is fixed soon, it is best to merge after beta for internal testing during the month.

LauraPoggies:
Error log pretty much fixed I think

ilkober:
As mentioned two times already, the country starts with 4 research slots instead of 3

Starting focus still mentions independence and features event about the resignation of the democratic leader even though it might not be the starting case of puppeted Quebec from a peace deal.

DEMOCRATIC BRANCH:
The tooltip about the "corruption scandal" reduction is still there in multiple colours. (mentioned twice too)
Union Nationale Wins focus features same leader tooltip twice.
"Expand Trade with Neighbours" still creates magical 15 steel out of nowhere.
Some decmoratic focuses require a tooltip of kind "if [FOCUS NAME] is completed: [SHOWCASE EFFECT]", check POL_devalue_zloty for more info.
"For the good of the nation" provides a national idea which is immediately removed if democrats are in power.

COMMUNIST BRANCH:
Collectivist tree still provides a total of -15% consumer goods, which is too OP.

FASCIST BRANCH:
"Contact the Blueshirts" still features an unclear tooltip about "some Blueshirt divisions", which was already suggested for rewriting.
"Abandon Ultramontanism" has a tooltip with a weird colouring which is unclear.
Purging clerics provides a research slot? How and why? Why not a sort of education reform?

LauraPoggies:
fixed above _except_ the research slots, setting it to 3 still keeps it at 4, considering adding a focus very early on in the economic tree that adds a very quick research slot but make them start at 2 slots

marijn211:
[20:38:36][texturehandler.cpp:160]: Texture Handler encountered missing texture file: gfx/leaders/Europe/Portrait_Europe_Generic_new_9.dds
[20:38:36][texturehandler.cpp:216]: Couldn't find texture file: 'gfx/leaders/Europe/Portrait_Europe_Generic_new_9.dds'
[20:38:36][texturehandler.cpp:160]: Texture Handler encountered missing texture file: gfx/leaders/r56_generic/portrait_commonwealth_generic_general_7.dds
[20:38:36][texturehandler.cpp:216]: Couldn't find texture file: 'gfx/leaders/r56_generic/portrait_commonwealth_generic_general_7.dds'
[20:38:37][texturehandler.cpp:160]: Texture Handler encountered missing texture file: gfx/leaders/Europe/r56_portrait_Africa_Female_Generic_land_1.dds
[20:38:37][texturehandler.cpp:216]: Couldn't find texture file: 'gfx/leaders/Europe/r56_portrait_Africa_Female_Generic_land_1.dds'

marijn211:
Some references to non-existent portraits need to be fixed, then it can be reviewed for next patch

LauraPoggies:
Fixed the missing portraits

marijn211:
The expansionist part of the political trees falls rather late, parts of the political branches before the conquering should probably be made optional so they could still be done later but don't delay it this much.
Economic focuses should also be buffed, probably by combining depression recovery and building up industry so there is not two separate branches to become capable that take up much time

By mid to late 1940 I was able to start demand stuff as communist Quebeq, which is only slightly late, but that was without being able to fix the depression debuff and with only 10 available factories.

marijn211 (changes requested):
See above, code-wise it's ready but it needs some balance tweaking

LauraPoggies:
Requested changes done, hopefully the balance is all good 🙏

marijn211 (approved):

marijn211:
I still feel like the economic branch, including tackle the economic issue could use more free factories, perhaps with less idea-based buffs, but it should be good enough to finally accept this first

PR #2641, opened 2023-06-09
Waffle june - I may not be able to open a PR later so here it is now

Thinking-waffle:
I may not be available for the update.
So here it is in advance.


marijn2112:
Pretty sure you confused < and > but I cant check it rn

ilkober (approved):
Looking good, man, looking good

PR #2644, opened 2023-06-14
Random fixes and tweaks for July update

hammmurabi:
just some random small tweaks to some trees
big things are:

focus reshuffle for portugal democratic path


common/national_focus/fake.txt
ilkober, line 1: Why is it there?

hammmurabi: from my understanding if you remove a focus thats in vanilla u have to put it in fake.txt, is that not correct? i removed the "protect chinese civillians" focus because it made absolutely no sense and was just cluttering the tree

common/national_focus/r56_manchukuo.txt
ilkober, line 1: One of them refers now to subject, the other one is free to take, why?

hammmurabi: prevent player from being locked out of entire tree if gains independence, for example from japan going democratic/communist

that path is weaker than the other so i made it be the one you can do without being a puppet

ilkober: No, there is no need to place it there
Just remove all the references to the focus if it comes up in our files and it usually should be fine

hammmurabi: ok

ilkober: For this (Manchukou), you could add a check for "having high relations (50/70+) with Japan" - that way a player can still access the tree if they gained independence in peaceful way

hammmurabi: hmm ok can do

ilkober (approved):

PR #2646, opened 2023-06-18
update characters from old style to new style

FOREVERGWC:
update characters from old style to new style，and I have tested it that is ok

### DIFF
--- a/common/characters/ADR.txt
+++ b/common/characters/ADR.txt
@@ -1,80 +1,80 @@
-characters={
-	ADR_co_prince_council={
-		name="Co-Prince Council"
-		portraits={
-			civilian={
-				large="gfx/leaders/ADR/r56_potrait_ADR_Co_Prince_Council.dds"
+characters = {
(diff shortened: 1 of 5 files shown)


marijn211 (changes requested):
The auto-indenters removal of many empty lines seems unnecessary, and probably makes the file less readable

FOREVERGWC:
> The auto-indenters removal of many empty lines seems unnecessary, and probably makes the file less readable
This is the standard result of vscode's paradox plugin's automatic formatting, where Shift+Alt+F ensures that everyone writes the exact same format

FOREVERGWC:
I recommend using the Paradox plugin of vscode for module creation. It automatically associates localized files to annotate key values, so many annotations are redundant and can quickly complete the code

FOREVERGWC:
@marijn2112  ok,you are right, I have restored the deleted blank lines.Which channel should I communicate this issue on Discord?

marijn211 (approved):

PR #2647, opened 2023-06-22
Update to new style and standard code

FOREVERGWC:
1.Programming style migration to new version (NSB DLC)
2.add simp Chinese yml


marijn211:
Chinese is not supported, a character loc file is of no use, a warning message could be written for Chinese though.
Most stuff seems aight although I didnt see where the removed Armenian and Azerbaijani OOB files were transformed

FOREVERGWC:
> The game now supports Chinese, and the latest version has added.
Armenian and Azerbaijani have not been used before. I conducted a global search for file names and templates, including the vanilla and 56. Because they use a new OOB loading method.
Such as:
image
image
In the common/decisions Armenian_decisions.txt and Azerbaijan_decisions.txt

marijn211:
I worded it wrongly, Road to 56 doesn't support Chinese, so its a bit pointless to add one file when the rest of the game will just be internal tokens (and someone also needs to maintain it, otherwise new characters will be missing very quickly)

FOREVERGWC:
> I worded it wrongly, Road to 56 doesn't support Chinese, so its a bit pointless to add one file when the rest of the game will just be internal tokens (and someone also needs to maintain it, otherwise new characters will be missing very quickly)

I am one of the creators of the 56 Chinese localization group, and I can maintain Chinese content. In fact, this is also the reason why I participated in Github.Everything always has a beginning, doesn't it?lol

image

PR #2654, opened 2023-06-28
New Partisan missions for occupied territory as germany

vini-44:
Adds a new mission that adds a state modifier in random occupied states depending on how many you occupy
Adds 2 decisions for each state to remove the modifier.
Modifier boosts resistance growth and generally is negative so the player is incentivized to remove it.
Ai has code to do the missions but might not be super balanced.


marijn211 (changes requested):
Seems very unbalanced and the way its scripted to block for AI also seems questionable

vini-44:
> Seems very unbalanced and the way its scripted to block for AI also seems questionable
how should i change it, rn im not sure how to rebalance it. the block is so the ai dosent try to do both negotianians and attack the partisans at once, also because they wouldnt move troops to the state to be able to do the decision. i'll look for a way to get the ai to do it right when i get the chance.

PR #2660, opened 2023-07-04
Updating Game Rules further

tichys:
Error was displayed regarding saudia arabia facist and democratic paths.

Updated pathing to correct path


tichys:
Ah and i also corrected the spelling from embodyment to embodiment

marijn211 (approved):

PR #2664, opened 2023-07-07
Coat of pain(t) for Philippines

Thinking-waffle:


marijn2112:
[01:14:53][pdxmeshtype.cpp:152]: Could not find mesh file gfx/models/units/planes/ITA_carrier_nav_3.mesh for generic_ITA_carrier_nav_3_mesh, switching to gfx/models/test_object.mesh
[01:14:53][pdx_anim.cpp:1125]: Animation [fighter_default_idle_animation]: Animation's number of joints does not match mesh's number of joints. ( Animation 1, Mesh 2)
[01:14:53][pdxmeshtype.cpp:324]: Animation [fighter_default_idle_animation] is not compatible with mesh [generic_ITA_carrier_nav_3_mesh]

marijn211 (approved):
Adds error log when BBA isnt enabled, but seems fine otherwise

marijn211 (changes requested):
On further inspection, it seems that the diff contains file changes that revert previous commits to the main (noticed on the gamerules file, but there might be more unintended changes)

marijn211:
Reverted presumed unintended changes to the gamerule file, except the removal of a recently added DLC check on the South African AI gamerule which is also checked for the Road to 56 tree, so I removed that check again

marijn211 (approved):

marijn211 (approved):

PR #2671, opened 2023-07-21
Brit august 1 

TheBrit7:


marijn211 (changes requested):
Use nudge to update unitstacks for the new province.

PR #2684, opened 2023-08-05
Beta bugfixes august 23

ilkober:


marijn211 (approved):
Idk what that Danzig event change means, but everything looks to be correct

PR #2687, opened 2023-08-06
Update POR - Portugal.txt

TheNordicG:
Fixed POR starting navy for non-MtG players


marijn211:
Can confirm this is an issue, btw, theres an excess bracket in history/units/POR_1936_naval_legacy.txt, if you have any further issues

For consistent organization, could you make this an else clause of the MtG ship variants, remember that it has to be inside the if statement and not after as that type of if/else doesn't work in history files

Also please merge branch again

PR #2695, opened 2023-08-15
small fixes 

hammmurabi:


marijn211:
Would you mind cleaning up the double availability blocks that seem to be present in some focuses?

hammmurabi:
> Would you mind cleaning up the double availability blocks that seem to be present in some focuses?

does it cause any issues? its like that due to the dynamic border war system i set up, from my experience it just merges them.

marijn211:
> > Would you mind cleaning up the double availability blocks that seem to be present in some focuses?
> 
> does it cause any issues? its like that due to the dynamic border war system i set up, from my experience it just merges them.

Nah just nicer organization, its optional

marijn211 (approved):

PR #2697, opened 2023-08-16
Uruguayan Focus

facundosim:


marijn211 (changes requested):
The election event thats supposed to bypass access focuses to all political branches isnt fired anywhere (I assume youd want to put that in an election on_action), also, "The American Revolution" has missing localization for the faction it creates

PR #2699, opened 2023-08-18
2K error goes brrr #RT56RP Original 

Simpyslav:
- This is a port-over re-enabling forgotten nation-specific icons LAMP included (Icons that existed in RT56 beforehand and with updates and DLC's became errors.
- Fixing the 2k error backlog plaguing RT56 for many weeks now
- Tested in the "official" RT56RP mod under duress in MP games and SP games to make sure it does not cause any issues.
- Please gib some credit to original fixer Warlider from the RT56RP Dev Team.
- Warranty not included for any issues that this might cause in THE FUTURE. However atleast error backlog is fixed.
- Best Simpy


marijn211:
Removing items or putting all the bugged ones to the same one pre-existing one is a drastic move, I hope someone from when these models were added could look into this?

Simpyslav:
The fix reverts most entities that didn't exist anymore to vanilla.
Other than that all original models that work + models that were just piling dust in the gfx folder were readded 
So the only issue that is visible is that all generic entities are available to everyone "Generic Heavy Tank = Copies French Heavy as an example"  However Nation Specific Icons will show up and priotize over anything else on the list.

Then again the 2k errors were left unchecked for months as no one has touched them so starting from the ground up with a decently working fix that adds absolutely forgotten icons back into the game is a positive way to re-add even more broken/forgotten gfx.

Some are adjusted by hand to not repeat however some might have leaked through.

Simpyslav:
@thinking_waffle and @abbus to my knowledge did most work in the folder  back in the dayso if they have any insight on their approach id love to hear it, The fix also helps FINDING more gfx errors in the future as it cleans up the errorlog

So TLDR:
Nation Specific Icons will be priotirized and work
Generic Icons get copied - Some where by hand adjusted not to repeat multiple times
Many forgotten icons are re-added so for example Japan can OI - Usa sees the beautiful T14 or the M6A1 
Gives RT56 its old flavour back instead of the same tank and airframes that were piling dust

marijn211 (changes requested):
Removes waffles cosmetic tag specific icons I believe, which suffer from false positive errors, tell me if theyve been relocated or changed but kept otherwise

marijn211:
gfx/interface/equipmentdesigner/graphic_db/00_plane_icons.txt is that false positive file btw, its all errors with `has_cosmetic_tag: Invalid Scope, supported: Country, provided: None`, same with 01_bba_plane_icons and `has_government: Invalid Scope, supported: Country, provided: None`

PR #2710, opened 2023-09-04
japan offset

Juddur:
should fix both dlc and non dlc offsets


marijn211 (changes requested):
Branches still dont shift to the left as they should, some even more to the right to line up, but the rest should move left instead

TheBrit7:
Is not fixed - closing.

PR #2715, opened 2023-09-06
Azerbaijan mod

Nowartus:
i hope i did the best


ilkober (changes requested):
Interesting first attempt but

Focus id's are incorrectly defined.
Many of the focuses are 140 days long
Most of the focuses are OP (5 factories for 140 days, first focus)
Localization and effects from Polish tree which have no use there.
Ideas are scaled off the top
Country-specific localization should be put in the same file

PR #2716, opened 2023-09-07
Azerbaijan

Nowartus:


ilkober (changes requested):
Completely lacks balance

Focuses seem very generic and not nation-specific

PR #2717, opened 2023-09-09
Fix focus "Foster Private Investment" for the Indian revolt.

TheycallmeBrick:
Based on focus requirements, it looks like this was meant to be Authoritarian or Fascist, rather than Authoritarian and Democratic.

### DIFF
--- a/common/national_focus/r56_india_shared.txt
+++ b/common/national_focus/r56_india_shared.txt
@@ -133,7 +133,7 @@ shared_focus = {
 		else = {
 			OR = {
 				has_government = neutrality
-				has_government = democratic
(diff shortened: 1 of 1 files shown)


marijn211:
The R56 tree also has democratic requirements though

TheycallmeBrick:
After the Azad Hind revolt occurs, they get a different tree, and the prerequisites change for their economic branches to simple government type checks. One is for Democratic and Communist, the other is for Authoritarian and Democratic. The problem is that Azad Hind only has a Fascist branch, so they don't have access to any economic branches. This is what that fixes, not the requirements for the standard India focus tree.

PR #2725, opened 2023-09-16
Azerbaijan Mod

Nowartus:
This was my last attempt. You told me you would do the balancing yourself. I hope I'm okay


ilkober:
There is no need for opening new PRs. This will not make them get accepted easier/earlier.

The current tree unfortunately is poorly balanced (e.g. -20% permanent consumer goods from oil exports, -25% stability or +30% war support from a single national spirit etc. ) and politically performs worse than a generic tree (including some Azerbaijani terrorism, what the hell?). Yes, it has names of famous people of a time in some focuses, but it doesn't provide solutions for the following questions:
1. What are the first priorities of the country after appearing on the map?
2. If USSR is fragmented, what path would/should they be able to choose?
3. How any of the political paths can be achieved?
4.  What interesting interactions with neighbours can the country do?


I would suggest instead of reopening the broken PR over and over while hoping for it to be magically fixed by us, to sit down and draw a plan of how and why each of the branches (economy, army, politics) should be structured and what should they feature.
Do some research, take some inspirations for mechanics from our other releasable trees (new ones are Ukraine and Quebec, for example), THINK AND PLAN AHEAD, focusing on plausibility and fun, then execute

Good luck, hoping to see some draft some day in #coding-discussion

ilkober:
I also never promised you anything about "balancing the designed tree" or "fixing bugs" for you.
All I said is that "it needs an extreme amount of polishing"
I never said that I will do it for you, so don't ascribe it to me smth

PR #2726, opened 2023-09-20
Azerbaijan Mod

Nowartus:
First of all, I am opening the new PR again after making changes based on your answers on my fork, and of course, I do not think that opening a new PR will be accepted any faster, and if you noticed, I have completely redesigned the focus.


ilkober (changes requested):
It still has decisions of promoting terrorism in Armenia, which is a huge no-no

ilkober:
My patience is running low
I will repeat one more time the following:

We have the dedicated channel on our discord called #coding-discussion, where inexperienced but enthusiastic modders like yourselves can get assistance and resolve some issues.

Right now the tree doesn't match any standards of our focus trees, so IT CANNOT BE INTEGRATED.
I wanted to play nice and suggest some reworks, but since you appear to not follow the pleasant approach, here is the number of point YOU will have to work out before even opening another PR (we can assist you in resolving some issues, but it can only be done in public discord channel, nowhere else):

1. There is no structure and focuses are scattered all over the place.
2. The tree uses up to no ideological checks, thereby nullifying the effects of the political tree.
3. Army, navy and air trees are worse then a generic one.
4. There is no diplomacy or foreign politics implemented.
5. No events are programmed.
6, No decisions are implemented.
7. Icon GFX are incorrectly defined and clod up the error log, some are also not showed at all.
8. Localization is broken and does not appear anywhere.
9. Some of the ideas are ridiculously OP and do not make any sense.
10. All focuses are very short and allow for about 3-4 years of gameplay - insufficient.

So. instead of opening the same (or slightly updated) PR over and over again, you have three options: (i) work on your own and implement all the mentioned changes in a hope that they fit ALL of our requirements, (ii) scrap the idea of doing a full tree revamp and instead create only political path(s) for the country, (iii) scrap the carcas you made, do some research, consult us and plan a new focus tree before attempting to implement it codewise.

We can provide constrcutive feedback and help you improve your content, but it has to have at least a concept behind it. Right now I see no concept whatsoever, but a bunch of arbitrary ideas mashed together in an attempt to reflect on some historical facts.

I wish you much luck and hope to not see another PR before you reflect on required changes.

PR #2727, opened 2023-09-21
Azerbaijan 

Nowartus:
I will keep correcting until you accept


Nowartus:
I may be inexperienced, but I will never give up on my enthusiasm.

ilkober:
That is great to hear
You can potentially learn faster if you just ask for advice in our public channel, as I suggested

PR #2729, opened 2023-09-21
Turkey Partial Rework

BiccyThiccy:
1: Fully reworked Army/Navy/Airforce Tree

2: Slight bug fixes + better AI

3: Some new decisions

Edits in response to feedback:

1: Added fully controls checks to fortification/naval-expansion decisions.

2: Fixed some slight issues


ilkober (commented):
Border fortification decisions should probably check of said states are fully controlled by Turkey - just to make sure you don't give forts to an enemy. Maybe also include a cancel condition where the decision will abort if state gets partially taken? 🤔

BiccyThiccy:
> Border fortification decisions should probably check of said states are fully controlled by Turkey - just to make sure you don't give forts to an enemy. Maybe also include a cancel condition where the decision will abort if state gets partially taken? 🤔

They already do. Look at the decision files, not the focus (the focus does nothing but unlock them)

BiccyThiccy:
I can add checks for the naval expansion focuses though,  in case that was what you were talking about

marijn2112:
Check error log

marijn211 (dismissed):
Check error log

BiccyThiccy:
Errors are fixed, currently working on the remainder of the localization.

marijn211:
I think the army tree might be better with some more extra buildings/free equipment mixed in, some more variety, something like Women in Aviation which you removed

BiccyThiccy:
> I think the army tree might be better with some more extra buildings/free equipment mixed in, some more variety, something like Women in Aviation which you removed

_Women in Aviation_ was replaced with a decision that unlocks after the first army focus.

I can mix in some free equipment effects, three of these focuses after all imply that their army modernization involves increasing arms imports.

Turkey was poor compared to most European countries. I think giving factories for army focuses would make their economy too powerful relative to its historical status. I already think this army rework is a net-buff. I'd be more interested in considering this after I rework the Economy/Industry branch.

I'll make a commit in the next few days, I'll ping you for a re-review when that happens. Thank you.

marijn211 (approved):

PR #2731, opened 2023-09-22
Update r56_baltic_shared.txt to add missing cores

TheycallmeBrick:
Added missing state cores when a player without WTT takes the Baltic Unification focus.


marijn211:
Looks correct, please update the fork and Ill accept this, idk if it makes much sense though to allow Estonia to form the baltics and Finland-Estonia, but then not allow them their focuses to stabilize the former union

marijn211 (approved):
The Estonia block needs some work, but the main fix takes precedence

PR #2733, opened 2023-09-25
Fascist India now has access to a Economic Tree Branch

Shizophrener:
Planning Commission makes the most sense, according to Free India Doctrine


marijn211 (approved):
Havent play-tested, but its small and directed enough that that shouldnt be a problem

PR #2737, opened 2023-10-01
Waffle october 3

Thinking-waffle:
The argeads are back.

It needs the cosmetic tag of the Macedonian empire to be visible (MAE_neutrality, MAE_fascism) and the Bftb greek tree. Check that there is no icon overlap (with continuous focuses for example). All because somebody complained about it on discord. "Things I do for love!"

Also more names, callsigns and operation names.


marijn211:
The legacy ID you remove seems to be referred to in R56_on_actions (and unique commander IDs)

marijn211 (approved):

PR #2744, opened 2023-10-10
Let the bugfixing begin

marijn211:


marijn211:
Note: if you have countries you think should only have 1910 trucks, you can remove the 1936 entries, I've tried to mass replace all history file entries of 1936 trucks to also have them start with 1910 trucks now those are a thing.

marijn211:
Btw Ill try to put Norwegian characters on here later
And adapt Finland's tree to the democratic starting ideology

TheBrit7 (approved):

PR #2745, opened 2023-10-11
Austria and Sudetenland state rework

TheycallmeBrick:
* Added 4 new cores states for Austria ("Burgenland", "Styria", "Carinthia", and "Salzburg"), and 2 for Czechoslovakia ("North" and "South Sudetenland"). Split necessary states for these new borders, added 6 new provinces in Austria, and updated all necessary unit, building, supply, and adjacency map data for these regions.

* Updated German, Austrian, Czechoslovak, Polish, Hungarian, and Romanian focuses, decisions, and/or events that would logically interact with these new states to include these new states.

* Added County of Tyrol (TYR) as a releasable tag in Austria and Italy, and added State 74 "East Sudetenland" as a core of Silesia (SIL) due to the region historically containing a significant Silesian population (i.e. Czech Silesia).


marijn211:
We clearly expressed our concerns about adding these states just for the sake of adding states, especially but not exclusively Burgenland.
Alongside that we have Czech content currently being made. Please consult team members before working on or opening more map PRs.

PR #2746, opened 2023-10-12
Japan hiding branch

Juddur:
Please check what i looks like in game, it is not done the exact way that vanilla does it. This is b/c vanilla's Manchuria branch is connected to a focus, hence why the offsets are set to go back otherwise the interconnect


marijn211:
Could you adjust this so everything moves about 3-4 focus widths left when purge kodoha is chosen (WtT owned)? Perfect otherwise

common/national_focus/japan.txt
marijn211, line 1: Now you broke it for people without WtT

marijn211 (changes requested):

marijn211 (approved):

PR #2748, opened 2023-10-12
Fix Danish States and Orkney Island

featherus:


marijn211:
I think giving North Schleswig its Danish spelling "Nordslesvig" would actually be more accurate than vanilla's term "Sønderjylland" which as far as I can find also includes the south

featherus:
> I think giving North Schleswig its Danish spelling "Nordslesvig" would actually be more accurate than vanilla's term "Sønderjylland" which as far as I can find also includes the south

Done

marijn211 (approved):

PR #2749, opened 2023-10-12
Fix unit models

featherus:


marijn211 (approved):
Hard to see if this would break anything else, but it fixes the issue and has no additional errors so seems fine

PR #2751, opened 2023-10-13
Monarchist Germany rework

TheBrit7:
Isn't quite finished yet, needs death event for Ludendorff and some potential bug fixing but other than that it's good to go. Opening the PR so it can be tested and reviewed accordingly.


marijn211:
FYI Im also making Germany stuff so Ill skip r56.ger.30 but I might use 31, 32 and 33

ilkober:
Adding starting cores for German pre-unification kingdoms from the start will open a giant can of worms. Better make it a decision/ game rule.

Simple example: USSR defeats Germany, can puppet Germany and release Prussia. Complete nonsense

TheBrit7:
Ludendorff still needs illness/death events so I'm going to need to take up those slots.

marijn211:
> Ludendorff still needs illness/death events so I'm going to need to take up those slots.

Maybe temporarily use a very high number like 999 then I can change it back to the next free number when I merge my stuff

TheBrit7:
Okay everything should be good to go.

ilkober (dismissed):
With Germany being a successor state of  Prussia, there is no need for both of their cores to exist from game start.

The existence of Alsace nation should probably fall into the Occitania gamerule about cores. Instead of having a starting core, their creation can be done through a decision (similar to how Lithuania can become Memel and vice versa)

Otherwise, everything else looks very promising.

TheBrit7:
> With Germany being a successor state of Prussia, there is no need for both of their cores to exist from game start.
> 
> The existence of Alsace nation should probably fall into the Occitania gamerule about cores. Instead of having a starting core, their creation can be done through a decision (similar to how Lithuania can become Memel and vice versa)
> 
> Otherwise, everything else looks very promising.

Did you even look at it? They're not there.

marijn211 (commented):
Ill have to replace my event IDs before accepting

marijn211 (commented):
Prussia cores were disabled, but theres still cores of other nations that should not be there at map start

marijn211 (approved):

PR #2760, opened 2023-10-25
Waffle october

Thinking-waffle:
manually uploaded my changes

test the new files for Iceland Québec, Sami and Finland

Implemented the new planes of the dlc (there may be some conflicts with the reskins I made but that will not cause problems anyway).

I need to add a "the" forgotten in the name of the focus "sideline council". 

Check that the sami character work (they should)


marijn211:
The history file of SMI is in the character folder

marijn211:
Theres another PR open thats nearly identical without the aforementioned issue, closing this one

PR #2762, opened 2023-10-26
Fix missing core for WTT German event.

TheycallmeBrick:
Added Vorarlberg to list of cored states when peacefully annexing same-faction Austria through event.


marijn211:
Already fixed on my branch

PR #2763, opened 2023-10-28
"Autonomous Soviet Republics" focus oversight fix

TheycallmeBrick:
Focus "Autonomous Soviet Republics" allowed SOV to puppet BLR and turn the nation communist, even if the tag was already released and a puppet of another nation.

Fix this by making SOV create a dynamic country instead if BLR already exists, achieving the same result but restricted to SOV's territory. The original code still applies if BLR doesn't exist.


marijn211:
Put a checkmark if you fix a bug in the bug channel, and if possible dont create a PR for a single bug, just hijack someone elses bugfixing branch (if repository access)

PR #2768, opened 2023-11-01
Pepper peru 5

pepper419:
-introduction of bop for peru between army and navy

-reform of the economic tree of Peru

-rebalancing of the equator tree from the war of '41

-rebalancing of some ideas from Peru


marijn211 (commented):
I still think you should remove division attack/defense from the BoP, maybe add some recovery rate instead (maybe (-)15 or 20% at most)
With naval too, would remove naval speed, hit chance and naval damage modifiers, not sure what it should be replaced with there, but your shells arent doing less damage because of morale or prestige

marijn211 (changes requested):
See above

history/states/1052-Acre.txt
ilkober, line 1: Places 30 rubber instead of 0

localisation/english/replace/r56_formables_l_english.yml
ilkober, line 1: Unifying Hispanoamerica as any other country buy Peru will make you Peru? Why?

ilkober (commented):
You added 30 rubber to Acre state???
0 -> 30

marijn211:
Pepper as long as this is marked draft I wont check this, just a reminder

marijn211:
Error log

marijn211:
Previous feedback also still applies, theres still combat stats in the BoP

PR #2769, opened 2023-11-02
Waffle Japanese shenanigans + MIO

Thinking-waffle:
Originally I just wanted to add the Japanese advisor pics.
But then I found a (slightly useless?) trait for the Japanese advisor to Nazi Gemany as an ideological crusader, find another Prince of Terror... and while I was at it added two focuses for the Kodoha to fight in Siberia.

Should be pretty reliable.


marijn211 (changes requested):
Diff is fucked

PR #2782, opened 2023-11-14
RT56 Turkey Improvements

BiccyThiccy:
**I would strongly suggest looking carefully at the Character changes I made.** I did my fair share of testing to make sure nothing's screwed up, but I am a novice modder who has never touched character changes before.

Changes:
- Most BFB vanilla tree advisors are available in RT56 Turkey.
- Some political focuses have had slight changes, primarily to accompany the changes to 'Sectarianism'
- Some duplicate advisors were commented out, these were causing errors before for non-DLC players.
- Requesting Hatay is now a decision, and the events associated with it have been partially relocalized, and have better effects.
- Several minor localization changes, both to my own stuff, and the old tree, which I think did a lackluster job in this area, at least by the standards we have now.
- Couple of minor changes to the army tree.

Eco changes (11/26 pushes)
- Fully reworked Eco tree.
- New decisions and spirits to go with it.


TheBrit7 (approved):

BiccyThiccy:
Brit mi amor danke so sehr 😀

PR #2784, opened 2023-11-22
Achievements and Ribbions

SpicyAlfredo:


SpicyAlfredo:
Waiting on icons

marijn211:
You could comment out the start_date_checks in the ribbons

marijn211:
Also lategame techs and late doctrines are basically just lame af waiting achievements

SpicyAlfredo:
> Also lategame techs and late doctrines are basically just lame af waiting achievements

That's the point of the ribbons, they're just passive feel good pop-ups. Not things you aim for or center a playthrough around.

PR #2787, opened 2023-11-30
a bit more

hammmurabi:


common/national_focus/mandate_of_palestine.txt
marijn2112, line 1: You have conditions as effects here

marijn211 (approved):

PR #2788, opened 2023-12-02
Changes to Bear Hunt branch of Soviet center tree, Restoration of Soviet "Great Commander" focus for Soviet center and Chinese unification gamerule

Idlemancer:
Made the focuses to give the Soviet Union different leaders after Stalin's death mutually exclusive with one another, adjusted their requirements and effects to be more appealing and made the focuses following them also more appealing

Also restored the "Great Commander" focus for Stalin's Inheritor of Lenin branch

Chinese unification gamerule at game start, options include variants for both Communists and Nationalists in either unification of warlords, total unification(including Japanese controlled land), total unification plus treaty ports, and the annexation of the Periphery of Tibet, Mongolia and Tannu Tuva


marijn211:
Please check error log before opening a PR next time, its not everything, but its the bare minimum of checks you should do

marijn211:
I like these changes but some feedback that you can optionally use:
-Maybe add cores for 1912 China starting scenario, the extra states arent very resource rich so to be a strength improvement that might be a good idea
-For making Voroshilov country leader, maybe require to be at war with a major? Seems like it'd make little sense for a military man to come to power otherwise.
-Maybe Qing China starting borders could be cool? Basically stealing 4 Soviet eastern states and Sakhalin, it could make a Soviet player maybe not have to worry as much about a Japanese invasion

marijn211 (approved):

PR #2792, opened 2023-12-05
Fix symbol errors

FOREVERGWC:
Fix symbol errors to solve the problem of translation tools not recognizing entries


SpicyAlfredo (approved):

marijn211 (approved):
Cant playtest fork, but looks correct

PR #2793, opened 2023-12-07
final small change israel

hammmurabi:


marijn211 (changes requested):
Story events like this are a bit inconsistent with other content, we think it should probably be in another form

hammmurabi:
> Story events like this are a bit inconsistent with other content, we think it should probably be in another form

its not a story event and i made it since the update gets released on hannukah i dont see the problem

hammmurabi:
its just a little flavour event just like greece has independence day and finland has the white death

PR #2795, opened 2023-12-09
Waffle mio january

Thinking-waffle:


marijn211:
Theres an error on Brazils focus tree where funds are added without it being specified to which MIO (from previous PR not this one)

Thinking-waffle:
Hello dear tester:

things you can try to test: - unify the celtic world as Scotland or Ireland and see which MIO you have access to.
Check CHI - PRC - Warlords - Manchuria - Menjiang (with and without wtt?)
Check Yugoslavia - Croatia (because inter country things are always something).


Normally that should be the end of the MIO integration without considering further addition such as: more companies, MIO for formable countries, things like that. Note that I may have forgotten a country too, you can check for that too.

Have fun, and a happy new year

SwedishNorwegian (approved):
No new errors from what I can gleam

PR #2797, opened 2023-12-12
SpicyAlfredo's Castle January 2024

### LINKED ISSUE #2776: AI seems to rush SRD research in early game while ignoring the basics
I've only played r56 for a little while but I've noticed, either when switching to other nations via console commands or just when I was unable to construct things like radar stations at almost any of my allies' territories without doing the faction research continuous focus (even around 1940 and when I had over a dozen of them).
This is because the vanilla research, especially things like the techs that give industrial/research buffs get kinda neglected by the AI, while there are often more than a half of their research slots going to S.R.D., which, if I understand this correctly, is meant to be rather used in the late game. 

Therefore I think there should be some (ahead of time I guess?) penalty in place that would decrease once the real essentials get researched, that would somewhat deter especially the nations with less research slots, from rushing SRD before covering the basics. Iirc some nations didn't have Electronic mechanical engineering while at least 2 techs from the SRD section were already researched.

SpicyAlfredo:
- AI Changes
- Tech stuff I forgot about ported from mod saves
- Tech AI etax "fixed"
- Number Changes, because changes = content
- Adjust historical France to make it harder for the AI to pilot (Buff to player?)
- China Rear fix 
- Random Finish focus added (Sapla Line)
- Tactic "blunder" only occurs for bad generals now. (moving combat along in general hopefully)


marijn211:
another_brick_in_the_wall_pt3has duplicate checks

SpicyAlfredo:
> another_brick_in_the_wall_pt3has duplicate checks

Fixed

marijn211:
Reverted GUI fix as Ive done this already on a different branch and you broke my fixing of the undersized nuke counter

marijn211 (approved):
Have not checked, please only use if well checked or important

PR #2798, opened 2023-12-14
heavy_fighter type changed to match vanilla

SwedishNorwegian:
changed heavy_fighter type from type = fighter to type = heavy_fighter

It is now possible to deploy heavy fighters once more with BBA.


marijn211 (approved):
Blind accept

PR #2803, opened 2023-12-16
Dutch hotfixes december

marijn211:


events/Afghanistan.txt
ilkober, line 1: Event about adding Pakistan into faction has to check that Afghanistan is in faction AND/OR faction leader.

Might be better replaced with decision which has an invitation event

events/Generic.txt
ilkober, line 1: This should add timed idea, not permanent one

ilkober: Even though it is an effect, it is not clear that it is timed one

localisation/english/AFG_l_english.yml
ilkober, line 1: AFG_kings_goon:0 "Ruthless Goon" - change localization string
AFG_reclaim_pakistan_mission:0 "Solidify Our Positions" -> change localization string

common/national_focus/afghanistan.txt
ilkober, line 1: Since big amount of focuses are now shorter (70 -> 56 days), are the historical focuses done during the historical timeline?

ilkober (changes requested):

ilkober (approved):

PR #2809, opened 2023-12-20
Heavy fighter bomber crash fix+Extra plane presets

marijn211:


marijn211:
Ive copied over the crash fix since Abbus hasn't committed his last changes, will close PR to let him decide whether to reopen already

PR #2818, opened 2023-12-28
Voice lines

hammmurabi:
voice lines for a ton of languages sourced from kaissredux + Original Hebrew voice lines


TheBrit7 (changes requested):
Let's not replace the vanilla voicelines, we're adding a load of storage space for a whole load of nothing. Some voicelines are also duplicates, they can be adjusted to multiple tags with the same voiceline files, check all the German Kingdoms.

In addition, we have to make sure some of the voicelines are up to quality. No offence to KX, but some of these don't sound vanilla-like, so a quality test is needed.

hammmurabi:
> Let's not replace the vanilla voicelines, we're adding a load of storage space for a whole load of nothing. Some voicelines are also duplicates, they can be adjusted to multiple tags with the same voiceline files, check all the German Kingdoms.
> 
> In addition, we have to make sure some of the voicelines are up to quality. No offence to KX, but some of these don't sound vanilla-like, so a quality test is needed. 

I'll go over it when i get home on monday

hammmurabi:
> Let's not replace the vanilla voicelines, we're adding a load of storage space for a whole load of nothing. Some voicelines are also duplicates, they can be adjusted to multiple tags with the same voiceline files, check all the German Kingdoms.
> 
> In addition, we have to make sure some of the voicelines are up to quality. No offence to KX, but some of these don't sound vanilla-like, so a quality test is needed.



fixed - removed the lower quality voice lines and the ones that were for nations who already had some in vanilla. Some files are "duplicate" with vanilla for example to add Italian to certain r56 releasables or German to the German kingdoms

TheBrit7 (approved):

TheBrit7:
Don't merge quite yet, still got some things to do.

PR #2819, opened 2023-12-28
Waffle january 3

Thinking-waffle:
-added effects to 00_on_action added by the last dlc. I would recommend to double check with winmerge or similar system to be sure that there are no other tweaks of the base game.
Those include the paratrooper sabotages, German trade with Sweden and something about auto-flipping states. Those states need to be checked as we may have added others in the same area, things like that.

-added extra CAN MIOs, CSA MIOs
-tried to fix the focus "CUB_winning_the_battle_of_the_caribbean", to test with and without MtG (it grants free early destroyer techs)


marijn2112:
Could you fix your previous PR using the wrong tech categories for special forces doctrines

marijn211 (approved):

PR #2823, opened 2024-01-01
Adding missing effects for Marine Commandos

Thinking-waffle:
Added 2 broken effects in Marine special forces doctrine tree
Changed some effects as we probably used effects back from the beta that have been edited since, especially marine "expeditionary task forces" - marines_expeditionary_support you can check if things are alright.

Added Soviet AAT effects. Warning it includes the extra special force tree which may be one tree extra with the generic decision.


marijn211 (approved):
Good, if someone else merges, make sure to update from beta if any other PR got merged already

PR #2825, opened 2024-01-06
January beta bugfixes

ilkober:


marijn211:
Why remove the MPL confirmation event

ilkober:
> Why remove the MPL confirmation event

It has no localization

marijn211 (approved):
Diff looks good, no errors, tested one of the leader descs and indeed didnt work
I hope you double checked all the removed presumably non-existent country flags

PR #2828, opened 2024-01-14
RT56 Turkey Changes: Diplo rework and political partial-rework

BiccyThiccy:
- The diplomacy section of the tree has been fully reworked. Plenty of brand new focuses have been made, and every single one of them is now (largely) independent of the political tree. New decisions and MANY events have been made to accompany these focuses too, so that there are more things to do with later-game political power.
- Most of the focuses below **Ataturk's Legacy** and **Support Anti-Kemalist Republicans** have been deleted, due to them being replaced by my newly added diplomacy branch. They have been replaced with new focuses, either related to the transition to democracy, or actions by the rival CHP and DP parties.
- Norman fixed something using this branch, idek what. If there's issues with anything GFX it's not me.
- A few thingies from BFTB that weren't disabled by using the RT56 tree have been modified to play better (Ataturk dying content, one MIO that's unusable in rt56 tur)
- The Soviet support focus is no longer necessary to begin the revolution. I think this makes sense, as the revolution still fires just fine without it. The AI will do the focus for it before the revolution happens.
- One bug from the army tree has been removed (focus trigger not working if you use SF or MW)
- The Venezuela focus **Demand Guyana** had a stupid bypass condition (look at the file change if you want). I fixed it to be much less illogical. The newest Colombia state also wasn't cored by the focus to anschluss them, this has been fixed.

**I have already done patch notes.**


BiccyThiccy:
I apologize for so much of this crap being on one branch. I hope the commit list doesn't look like the Vic2 late game industry screen

BiccyThiccy:
Me, Fantom, and to a lesser extent Norrman, took a look through my TUR focus changes over Discord. I've implemented suggestions based on them just before Fantom went to bed.

I don't think either will be looking at my PR directly, but as far as I know they're fine with everything, at least until I get to doing the political section of the rework.

If there are small things you want changed, feel free to do them yourself if you have a concrete idea what you want to do. I am busy-ish, but I should be able to implement changes based on feedback before we get to Jan-1.

marijn211 (approved):

BiccyThiccy:
:D

PR #2829, opened 2024-01-15
Scottish stuff mostly

Thinking-waffle:
Some effects may be a bit too hard to do 100% correctly we will see (mostly diplomatic interaction especially regarding Bavaria)


Thinking-waffle:
Have fun playing Scotland, please give feedback.

Oh I still need to add the stone of scone thing, oh well...

TheBrit7 (approved):
SCOTLAND FOOOREEEVER

PR #2831, opened 2024-01-18
Quebec bugfixes & Advisors

LauraPoggies:


ilkober (commented):
Mentioned issues are to found in Laura's channel

LauraPoggies:
That should be all the issues, hopefully.

marijn211:
In what contexts do the accented VP and state names cause issues?

marijn211 (dismissed):

marijn211:
[23:29:57][no_game_date][effectimplementation.cpp:2152]: common/national_focus/quebec.txt:7301: swap_ideas: Invalid idea: QBC_quebecois_socialist_agrarianism. If you wanted to reference a variable, you need to explicitly prefix it with var: or temp_var:.
[23:29:57][no_game_date][effectimplementation.cpp:2152]: common/national_focus/quebec.txt:7370: swap_ideas: Invalid idea: QBC_quebecois_socialist_agrarianism. If you wanted to reference a variable, you need to explicitly prefix it with var: or temp_var:.
[23:30:02][no_game_date][lexer.cpp:213]: Tried to get dynamic token that does not exist: civilian_cost_factor

marijn211 (approved):

PR #2834, opened 2024-01-20
Waffle february 2

Thinking-waffle:
This adds a secret focus to unlock a special operation to get the Stone of Scone back with agents. It is necessary to test it.
To unlock it the English must refuse and then instead of getting a wargoal if you have the dlc you can focus on spies instead and gain a secret focus.


marijn211:
It doesnt make sense that free trade has a smaller buff than export focus, it exports more resources
Also, it seems more sensible to put such buffs onto the conscription laws, as otherwise countries with no resources to export get even more buffs for free.

marijn211 (dismissed):
[00:24:17][no_game_date][operationphase.cpp:66]: Operation phases references a phase that does not exist: targeted_take_the_stone_of_scone

Thinking-waffle:
I see what I missed, I could have variety but that's enough for now.

It should be fixed (I changed the text too as it was incorrectly set up with the description used as the title)

marijn211:
Ill took a look at the vanilla balance changes you ported and accept if those seem right

marijn211 (approved):

PR #2838, opened 2024-01-27
Attempt to boost Coordinated Strike

Thinking-waffle:
This is very experimental to try to boost Coordinated Strike spy mission. One of the hardest missions with a very low payoff.

Here it gives a 10 day boost of +10% attack against the victim and +15% breakthrough against the victim (if I did the scope correctly).

I also doubled the length of the no interception duration, doubling the number of special attacks then. If that system works ofc.

It needs testing. I used a flag in the operation file to try to target the victim correctly. But the correspondence between the target of the operation and the target of the attack bonus provided may not correspond, that's what needs testing.


marijn211:
[14:42:54][no_game_date][effect.cpp:422]: Invalid scope type for effect add_intel in common/operations/00_operations.txt line : 2024
[14:42:54][no_game_date][effect.cpp:422]: Invalid scope type for effect add_intel in common/operations/00_operations.txt line : 2028
[14:42:54][no_game_date][effect.cpp:422]: Invalid scope type for effect add_intel in common/operations/00_operations.txt line : 2032
[14:42:54][no_game_date][effect.cpp:422]: Invalid scope type for effect add_intel in common/operations/00_operations.txt line : 2064
[14:42:54][no_game_date][effect.cpp:422]: Invalid scope type for effect add_intel in common/operations/00_operations.txt line : 2068
[14:42:54][no_game_date][effect.cpp:422]: Invalid scope type for effect add_intel in common/operations/00_operations.txt line : 2072

marijn211 (commented):
Idea doesnt correctly target the operation target, intel bonuses are a bit pointless since you require a strong network to do the operation

marijn211:
Except what I commented everything should work, but waffle can adjust/remove the broken part when hes available

PR #2842, opened 2024-02-02
Bugfixes for Beta

ilkober:
Fixed a misspelled tooltip for Portuguese MIO.
Corrected the arrest of Raghib Al-Nashashbi not working properly.


common/national_focus/portugal.txt
marijn211, line 1: Why would you replace the tooltip of the idea with a tooltip of an MIO that might not even be loaded?

ilkober: por_ogme does not exist, hence why I just picked the other available option

marijn211 (approved):
Looks good, testing one more bugfix then accepting

PR #2843, opened 2024-02-03
Bag of Tricks Feb

SpicyAlfredo:
- Reduce Stat bloat and Vanilla balance adjustments
- Asymmetrical Doctrine more "Vanilla like" in balancing 
- Adjustments and polish in various files


marijn211:
Error log

marijn2112:
I fixed the Peru war with Ecuador being duplicate, and then the issue with them joining factions started happening, so instead of the AI priorities, there non-intervention ideas are probably not being added

marijn2112:
The victory point defines included now are buggy, I previously had to remove them from PRs by Brit

marijn211 (changes requested):
See above

marijn211 (approved):

marijn211 (commented):
Gibraltar strait is still always friendly

PR #2847, opened 2024-02-06
German Civil War: Player retains GER tag

IyeOnline:
* In the german three-way civil war, the player now retains the GER tag
* First, two dynamic countries for the uprisings are spawned in order to allow the player to see the distribution on the map
* When the player makes their choice in the event, territories are changed around between the dynamic countries and GER, so that the player retains the GER tag
* The death of hitler event upon capturing berlin from the monarchist civil war will now trigger in the threeway civil war. Its effect and text are changed to correctly reflect that.
* There are three new cosmetic tags to ensure that the colors retain consistent and logical throughout this. I.e. fascist germain retains the black color, while communist germany gets a red color and democratic germany gets a blue-ish color
* Besides those changes, the actual civil war (territories, troops, ...) remains unchanged.


SwedishNorwegian:
- "The Death of Hitler Event" This event should not fire upon the capture of Berlin as you kill him in the focus leading up to the civil war.

IyeOnline:
Its no longer the "Death of Hitler" event. Its now the "Death of the Führer" and it also correctly tells you who the new Führer is.

It kills the current fascist leader and follows a succession similar to the one from the "Kill Hitler" event. I.e. you go Hitler -> Himmler -> Göring or Göring -> Himmler depending on where you started. For the monarchists nothing changes.

---

I triggered it during a test run and though that giving it a bit of flavour instead of disabling it would be nice. The only thing is that the event now speaks of "the Führer" as the dead guy, since that was a easier than figuring out the name of the leader before he gets killed.

marijn211:
Please resolve the conflict and update from the main, then Ill download the fork to playtest

marijn2112:
GER_scripted_effect needs state 1053 included in `GER_civil_war_make_this_democratic_preview`

marijn211 (dismissed):
- GER_scripted_effect needs state 1053 included in `GER_civil_war_make_this_democratic_preview`
- Bormann can't be made leader, he has no portrait (you can discard that part of the if/else statement, and also combine the code where you replace Hitler and Göring, which use the same effect)
- Death of Fuhrer event should mention the old leaders name and it shouldnt fire as quickly for the 3-sided civil war (probably include capitulation progress) since rushing Berlin can be a bit too easy

(theres also an empty, unused scripted trigger)

marijn211 (approved):

PR #2848, opened 2024-02-07
Bugfixing

### LINKED ISSUE #2851: Multiplayer crash due to file name of bulgarian navy
I wanted to play with my friend (he on Windows, me on Linux) and the game crashed almost directly with the following message:

`[18:37:51][1936.01.01.12][virtualfilesystem.cpp:253]: Could not open file: history/units/BUL_1936_naval_MTG.txt, error: not found`

Instead of _"BUL_1936_naval_MTG.txt"_ the file was called _"BUL_1936_naval_mtg.txt"_ with lower case "mtg". This might only be a problem for Linux users, as Linux handles files case sensitively while Windows doesn't. 
We solved the problem locally by both renaming the file to _"BUL_1936_naval_MTG.txt"_ (we both needed to rename it, as changing a files name also changes the checksum).

Also, in the current branch there are two files _"BUL_1936_naval_MTG.txt"_ and _"BUL_1936_naval_mtg.txt"_ with the exact same content. Might this be linked to this problem?

marijn211:


marijn211:
@ilkober error log

marijn211:
Scrolled through the diff which looks good, error log is clear

SwedishNorwegian (approved):

PR #2849, opened 2024-02-08
rename moselland to rheinpfalz

ItsFlor:
just a quick thing, it it doesnt really matter. honestly up to whoever merges this


marijn211:
Maps show the area being called Rheinpfalz consists only of the southernmost 40%~, while Mosselland is accurate for another 40-45% of the state, since both are around equal in accuracy, I think its sensible to stick to the current name

PR #2853, opened 2024-02-15
Bugfixes 2.0 + Tweaks for smoother transition with ToA release

ilkober:


marijn211:
Personally wouldn't remove those historical scripted triggers, because its a lot more precise for event option AI for example than just doing is_historical_ai = yes or however thats named

SwedishNorwegian (approved):

PR #2862, opened 2024-03-02
March bugfixes+TOA porting

marijn211:


marijn211 (commented):
Fantom incorrectly instanced Latvian characters, wait with accepting

PR #2864, opened 2024-03-05
Fix province 755 wrong name, Added historical Brittany

Altitud49:
Province 755 has been given the name "Brive-la-Gaillarde" so it does not show the state name "Limousin" anymore as it is since 2023; a victory point.
Can be renamed to Sarlat-la-Canéda or Villefranche-de-Rouergue if geographical accuracy is preferred over regional importance though having a at the time 6,000 inhabitants town being prioritised over a 30k one seems a tad ludicrous.
Could also be renamed to simply "Brive", its usual abbreviation to make it a little bit more bearable to the eyes on the map. _Minor Bugfix_


Altitud49:
Brittany can be split into Marche-de-Bretagne and Finistère if there is demand for it, due to the "infrastructure problem". Historically Rennes and Nantes had a similar infrastructure to those of Cholet, a little city in neighoring Loire state, a level 4 infrastructure state. This is not comparable to Lannion or even Pontivy whose infrastructure was... well let's just say it was at similar to the one of Centre-Sud. Alternatively, all of Brittany could be turned into dogwaterzone by putting the infrastructure back to 3 even with Loire-Atlantique/Inférieure added.

marijn211 (commented):
Based on what did you remove like 2 million in population

Altitud49:
Thank you for pointing this out, it appears I do not know how to use a calculator. Population transfer comes from Loire-Atlantique being transferred from Loire to Brittany. [Data from Wikipedia](https://en.wikipedia.org/wiki/Loire-Atlantique#Demographics). Should now be okay _(Edit : Data from the original game seems to be correct but rounded to the nearest hundred, may be changed for consistency but really doesn't matter)_

PR #2866, opened 2024-03-08
Province 755 stops getting called generic "Limousin", historical Brittany, vp moving for geographical correctness

Altitud49:
With all your favorite Tibetan changes integrated you can now enjoy your truly amazing Brive-la-Gaillarde, historical Brittany, and vp adjustments in north-western France !


Altitud49:
Now that I tink about it, instead of doing injustice by naming a vp "Villefranche-de-Rouergue", arguably less regionnally important than "Brive-la-Gaillarde", or having to live with a not geographically correct vp, I resolved the status quo by moving Brive on top of Brive ! 👍

Altitud49:
Returning to infrastructure consistency and balance what do you think should be done ? Bretagne_infrastructure
In the undermentioned map we can see that the vanilla province map doesn't include large parts of eastern **Pays Rennais**, which correspond to province 526 and smaller parts of **Pays Nantais**, this will be changed in the future.
Blue and Red lines show areas that were slowly chipped-away by the Franks from the Breton Celts, area between is called Marche-de-Bretagne (literrally March of Brittany or Breton March). Cyan line divides **Cornouaille** into Western Cornouaille, which forms "Finistère" with **Léon**, and Eastern Cornouaille which could be considered the "lowest infrastructure level area" of Brittany. 
Bretagne_pays
Compare to changes for now including Nantes and Saint-Nazaire provinces which are approximately Loire-Atlantique.
Bretagne_provinces
Additionnaly Loire-Atlantique/Pays Nantais/Bro-Naoned wouldn't be a core but a claim of Brittany in 1936 but there could be a decision or focus once you control it ?
Anyways what do you think ?

marijn211:
Our Belgian France expert sais it is sensible to have the tiles like this, although Vichy in WW2 mightve messed around with the organization and removed Nantes even before it officially happened in 1955

marijn211:
I dont know how big these new VP cities are, but you doubling the amount of VPs in the state makes me wonder whether you're adding smaller cities than are omitted elsewhere

Altitud49:
The thing is, if Brittany ever got its independance, it would be with Nantes. I just feel like administrative divisions of a puppet state that doesn't exist at game start in a territory they never historically controlled as "states" that define borders is a bit weird.

Altitud49:
Hopefully, new vp distribution is now good, Nantes has been put back to 5 (I dunno why I set it as 7), Saint-Nazaire went from 2 to 1 (biggest shipyard in France since the 1900s). As for Saint-Brieuc and Saint-Malo. They have been relegated to Lyck status, 0 vps but still in the files for train connections or supply mapmode.

PR #2867, opened 2024-03-09
Victory point adjusting, Brittany existing

Altitud49:
For the last time I hope, this fork has been recreated, changes range from simple new color for Brittany based on EU4 and other mod Kaiserreich/Kaiserredux to province rework and state rework to allow for historical Brittany to get independance, on top of that, victory points in western France have been readjusted in some places which involves making Rennes Brittany's capital, as it always has been, adding glorious Caen, major city very close to the Normandy landing beaches and more!


SpicyAlfredo (commented):
Looks good, don't like the VP values of 7 and 12 as I prefer to stick to 1-3-5-10 etc. Will probably approve once tested.

PR #2868, opened 2024-03-10
Carrier aircraft UI update

theodorhansson:
When building aircraft carriers and setting its default aircraft complement, one could see extra airframes (that are unbuildable). This update removes these superfluous items.

This update also standardizes the names for CV aircraft in the UI mentioned above along with the "quick-deploy" window.


AbbusF (approved):
Thanks for the work, appreciated!

Checked both with and without BBA.

PR #2869, opened 2024-03-10
🥰 Biccy April 2024

BiccyThiccy:
- Bug with the Peru-Bolivia formable resolved
- In progress: Reworked the Persian Empire formable from the ground up, changing it from an all-or-nothing formable to one where you can expand and integrate crap at a much more plausible rate.
- To-do: Rework the Iran Army Tree and add focuses to the Eco
- To-do: Ottomans should get more advisors


ilkober:
Not a bugfixing branch? Okay

PR #2874, opened 2024-03-13
Make AA rocket rail only heavy fighters work

Dentyr:
Previously if you tried to make an AA rocket rail only heavy fighter, it just wouldn't work because the game couldn't recognize the equipment type.


SpicyAlfredo (approved):
Seems good

PR #2878, opened 2024-03-16
India Fix (Variable Renaming)

Shizophrener:
Just changes the state variables `inc` `upa` and `ifb` to have an _state appended. Fixes the elections not working.


marijn211:
You made a typo in one of your ifb_state replacements, also why would the old names not work I wonder?

Shizophrener:
I assume they added some internal usage for inc

marijn211 (approved):
Ive confirmed theres no old versions of these variables in the mod

PR #2880, opened 2024-03-17
Fixed CHI Hanyang Arsenal Tanks not unlocking and RAJ John Mathai not unlocking upon completed focuses

picturetheory:
Hanyang Arsenal Tanks wasn't included in MIO file, typo in Mathai's requirements.


marijn211 (changes requested):
The available check for RAJ is not incorrect, instead the focus check should only show for the vanilla tree, while the country_flag unlocks it for the R56 tree

Either remove the edit, or replace it with an if/else clause so only the applicable triggers show for the correct tree

Also use show_mio_tooltip = my_mio with DLC checks for the hanyang unlock focus

marijn211 (approved):

marijn211 (approved):

PR #2886, opened 2024-03-19
More accurate Yalta+Post-war stuff

marijn211:
I've removed the British-Ethiopian agreement which only make sense to put on a larger PR when testing is more needed anyway.

- [x] Yalta adjusted: No instant WGR/DDR, and some fixes
- [ ] Decisions to create WGR/DDR  (halve-finished)
- [x] Non-Yalta occupation decisions for Iran, also fixed multiple bugs with operation countenance


ilkober (approved):
In errorlog we trust

PR #2888, opened 2024-03-22
niger state rework

hammmurabi:
shifted some provinces in the niger state to make fighting in french africa more possible

also fixed austria airports


marijn211 (commented):
Is the population increase intentional, everything else looks good

hammmurabi:
yes, https://en.wikipedia.org/wiki/Colony_of_Niger this lists the population in 1936 as 1,747,000

marijn211 (approved):

PR #2889, opened 2024-03-22
added alternate ruler for communist germany

ItsFlor:
If germany doesnt ally the USSR when going communist, Heinrich Brandler can take over the party and become the new country leader. germany gets a new colour and name.

added portrait, slightly buffed the soviet alliance to compensate for the leader buffs


marijn211:
Personally dont like the idea of replacing a leader you just fought a civil war for (also biased cuz Thalmann gets the poggers uniform change)

Also events should stick to the a/b/c standard

ilkober (dismissed):
An event should provide options and not just forcefully change your leader.

Also, what is the colour change for?

common/characters/GER.txt
ilkober, line 1: Character name should be stored in Characters file

ItsFlor: Other option is the mutually exclusive focus for leader
colour change is like trotzkyist ussr, they get a different colour too

ilkober: Event should make you be able to select a leader, not just forcing the change. Especially in context of potential USSR civil war and alignment of that kind

ItsFlor: added an additional option now

ItsFlor:
-updated tooltips for several focuses, making them more readable
-fixed issue with treaty of versailles path not being able to do the last focus
-added new name for communist germany, changes with alliance with the ussr or with brandler
-updated some focus icons to look more communist

marijn211 (dismissed):
Error log

SwedishNorwegian (approved):
LGTM

PR #2891, opened 2024-03-29
German States

HBN-Trader:
**I am working on fleshing out Germany.**

I have added four new states:

- Grand Duchy of Oldenburg (occupying the Weser-Ems state)
- United Thuringian Duchies (a confederation of the Ernestine duchies and the Reuss and Schwarzburg principalities represented by the Duke of Saxe-Meiningen)
- Grand Duchy of Brabant (Flanders and Noord-Brabant, ruled by an Arenberg prince, supposed to be a German puppet formable if Belgium and the Netherlands are conquered)
- a Sudeten German state (will eventually be formable as a third alternative between war and annexation of Sudetenland as well as possibly in an independent Event chain where German nationalists try to break off Czechoslovakia without German support)

They are not formable through the Occupied Territories tab yet and I am going to add decisions for them.

I also fixed the state name of Niederösterreich.


marijn211 (changes requested):
You've messed up some stuff it seems as the game rule file shows the reversal of DLC patch changes.

Also, Sudetia and Brabant at least sound more like larpy map painting things than making any real sense, it hurts me as a Brabantian, but like 50%+ of the included land was never in Brabant

You should probably discuss the merits of your content with us on the Discord

HBN-Trader:
> You've messed up some stuff it seems as the game rule file shows the reversal of DLC patch changes.

I'll try to fix that tomorrow. Looks like I forgot to download the beta updates.



> Also, Sudetia and Brabant at least sound more like larpy map painting things than making any real sense, it hurts me as a Brabantian, but like 50%+ of the included land was never in Brabant
> 
> You should probably discuss the merits of your content with us on the Discord

Sudetia is supposed to be a third way option between annexation by Germany and outright war, also maybe voluntarily giving them autonomy (i.e. creating them as a puppet) can be a way for Czechoslovakia to reduce tensions in that region.

Brabant (and perhaps 1-2 other states I will create in the Benelux region) is indeed ahistorical, but I imagine that this is what Germany might do if it occupies Belgium and the Netherlands, to reduce local nationalism and to secure the loyalty of Flemish nationalists without strengthening the Netherlands. Brabant would become a standard German "Reichsland" just like Bavaria or Saxony eventually. Germany had all sorts of plans regarding the establishment of new monarchies in occupied territories during WW1 and this is how it might play out.

The game map's borders are unrealistic. I am aware that Brabant includes large parts of what is in fact Limburg, for example. Maybe Germany decided to consolidate Flanders together with Dutch catholic territories (i.e. North Brabant and Limburg)? This would also be a major concession from protestant Prussia to Bavaria (allowing the formation of a new majority Catholic monarchy).

Oldenburg is also bigger than in real life...it includes a strip belonging to Hannover at the Dutch border. I can imagine that in a scenario where Germany restores the 1871 constitution, Oldenburg gets this and Münster in return for not getting back the exclave in Holstein and Birkenfeld, as the German government would certainly use the restoration of the monarchies as an opportunity to clean up some of the messy internal bordergore.

ilkober:
Majority of those additions is monarchist larp for the sake of monarchist larp.
I am yet to see sources on what it was based upon.
Effort in general is appreciated, but do some research first before adding random rp states on the map. Thank you

PR #2896, opened 2024-04-03
Possible Fix for newest Error Log pop up with some QOL improvements.

Simpyslav:
Please let me know if this is acceptable - some inconsitency may be visible because of the recent changes it's quite hellish to figure out why its been throwing an error all of a sudden.

(Not a single patch-note mentioned this shadow change from the official hoi4 dev side)
There's a few minor nation fixes that showed up as errors late into the game - based on 4 games run to 1956.
Rest is mostly AI strategy issues.


marijn211 (commented):
Diff looks good, so if the error log is all good Ill accept in a moment

marijn211 (approved):
Should be a net improvement, despite there still being a lot of broken references in the mod

PR #2901, opened 2024-04-08
Bulgaria: removed duplicate definitions file

theodorhansson:
The file \history\units\BUL_1936_naval_mtg.txt is defined twice. Since Windows path does not handle cases, it is best to remove this duplicate.


marijn211 (approved):
Content seems identical

PR #2902, opened 2024-04-08
Cringe meme deletion

theodorhansson:
Split from PR#2858

There exists some memes in the locale. This PR removes these.


SpicyAlfredo:
The ship ones are fine changes, some of the rocket ones aren't really "memes." Don't change those and add desc for the rocket ones that are memes e.a Harmbe and the french one and I'll accept the pull.

theodorhansson:
@SpicyAlfredo I restored the ICBM locales (without baguettes or monkeys).

marijn211:
I assume Spicy is forgetting that the rocket things are overrides of the default loc, and that the default non-meme loc will be used if they are removed

marijn211 (approved):

PR #2904, opened 2024-04-13
Waffle may 1

Thinking-waffle:
important:  

- test if the new mongol, tuvan infantry using PRC models with soviet rifles and smgs work.
- test PER tanks (for now without NSB) (it should be CZE tanks except early light tanks with a Czech made Romanian tankette)


marijn211 (dismissed):
SWI file is incorrectly instanced, Ill fix that at some point

marijn211:
Error log free on launch, so the models spoken off can't be too broken

ilkober:
There are currently 15 conflicting files...

Thinking-waffle:
oh boy I probably won't be able to fix such a merge today.

marijn211 (dismissed):
Doing final checks rn

marijn211 (changes requested):
Crashes for unknown reason after merge, am checking

marijn211 (approved):

PR #2906, opened 2024-04-20
AI Design Adjustments + More

SwedishNorwegian:
- AI will now design better Tanks
- AI will now design better Planes
- AI Majors will now try and research Medium Tanks earlier
- AI Tank divisions will now aim for 30 Width
- AI Nations have been put into categories which instructs them on what to research earlier and what to ignore
- 1936 Light Tank Chassi has been moved to 1937 to increase the value of 1934 Light Tank
- Removed Conversion Technology from Industry and baked it into Machine Tools
- Total Conversion Technology Modifier has gone down from 80% to 45%
- Removed Volkswagen and Deutsche Arbeitsfront ideas and merged into Volksgemeinschaft to reduce Idea Icon Bloat on Germany
- Added Idea that prevents Vichy France from rejoining the Axis when Case Anton happens.
- Readjusted the IC cost of Small, Medium and Large Airframes as well as the different Engine Types. This is to prevent the insane air wars that end up happening with production scaling way out of hand.
- Italy will now abandon its hard earned colony in Eastern Africa, much to the rejoice of the +250k Manpower it usually sends to their deaths there.
- AI will now prioritize Air and Tank research over Extended Technologies if they are a major up until 1940.1.1
- Fixed invalid design for US when it comes to M4 Zippo and M3A1 Satan tanks.
- Moved over Greater Ukraine formable from Maps-March branch to this one as well


marijn211 (commented):
I think it makes more sense to just put factory checks for the strength categories than tags, you can always add exceptions to the categories when a major is weak at the start

A strong Chinese warlord should not end up in the same category as Andorra or an African releasable with no factories

This does mean the checks can no longer be in allowed which makes it more performance-intensive, but the current triggers are quite arbitrary especially with strongly alt-hist games
I also don't think you're losing any AI intelligence as most of the times, it'll come down to the same thing

SwedishNorwegian:
> I think it makes more sense to just put factory checks for the strength categories than tags, you can always add exceptions to the categories when a major is weak at the start

The US for example matches the requirements pretty early in the game, usually around 1937 as the AI looks at its economy and sees that it's got a huge discrepancy in Civs to Mils and then only builds Mils for quite a long time, the French and British usually get going with Basic Mediums in around early 1937 which is fine.

> A strong Chinese warlord should not end up in the same category as Andorra or an African releasable with no factories

All Chinese tags are forced into the No-Tank-Research category because they are in no position 9 times out of 10 to actually build tanks. Thus by taking that research away from them they focus on other things first.

> This does mean the checks can no longer be in allowed which makes it more performance-intensive, but the current triggers are quite arbitrary especially with strongly alt-hist games I also don't think you're losing any AI intelligence as most of the times, it'll come down to the same thing

The tags that are relegated to the Minor-Minor category should never build tanks anyway. The categories are fine and those who are relegated to Minor category will instead of diverting research to Medium tanks which require more industry than they generally have will focus on Light Tanks which you can research earlier and are cheaper to build.

marijn211:
The removal of two engine on large frame 0 fixes an inconsistency with the other large frames, but putting just 2 engines does seem viable especially with light equipment like naval recon, vanilla also doesnt allow it though so Im indifferent

marijn211 (commented):
Everything Ive checked so far looks good, would only do a confirming AI comparison [do countries of all strengths have more sensible research+what does the plane rebalance affect] and maybe compare light/medium planes
Inter-engine balance seems good, I didn't realize engine 2->3 had such a big jump,

ilkober (commented):
Only checked decisions, focuses and bugfixes - look solid to me.

AI or templates left for @marijn2112, otherwise will approve changes

ilkober (approved):

PR #2908, opened 2024-04-28
israel independence war changes

hammmurabi:
some changes to israel war of independence

1936 arab revolt now happens on the correct date (april 19th 1936)
Israel will no longer start the war with control of beersheba
fixed the local offensive modifiers so they will only apply to the correct country (also for afg since it uses the same system)
MPL no longer has a say in the king edward abdication crisis for some reason
fixed iraq getting ceasefire event even if not in the war


marijn211 (commented):
Arent the ROOT scopes in ai_strategy unnecessary? Theres a few exceptions where it isnt default but thats rather rare

hammmurabi:
> Arent the ROOT scopes in ai_strategy unnecessary? Theres a few exceptions where it isnt default but thats rather rare

I put them in to try to fix a bug but it wasn't that in the end and didn't think it was too important to remove them.

marijn211 (approved):

PR #2911, opened 2024-04-30
Characters for ELS and SAN

facundosim:


ilkober (changes requested):
The brief check indicates that some of the Salvadorean advisors were born in 1920 or 1925, which shouldn't make them army chiefs, military theorists or high command stuff by the start of the game, but probably rather later. Please doulbe-check their age/profession by 1936 andd change them accordingly.

Apropos, we have a trait "promising", which could be used for those officers who are very young/inexperienced, but can learn fast/rise fast to higher ranks.

ilkober:
Will keep closed until requested changes are made

PR #2916, opened 2024-05-03
Beta Bugfixes

ilkober:
Drop the bugfixes, accept late


marijn211 (approved):
Ive checked this PR since its your own, everything seems okay, might double check the events with trigger removed if I have time

PR #2918, opened 2024-05-05
Upkeep and Minor Changes

Simpyslav:
- Reshuffled Paraguayan Tree to fit more sensically and additional focus for sprinkle Lore ;P
- Generic Events
- MIO Additions
- Upkeep of older trees, modifying effects
- Uruguay now targets "Axis" for loans, and requires their acceptance
- Bolivia Decisions now Overwritten,fixing the decisions targetting wrong states (Don't think Bolivia will have any more additional decisions added so it's not a big loss)

Saw Paraguayan Errors, Left Temporarily to see the opinions of others on the matter


ilkober:
Outdated branch, needs an update and some newer tweaks

PR #2920, opened 2024-05-07
Fixing two mistakes in define and CZE History

SpicyAlfredo:


ilkober (approved):
mistakes were made...

PR #2921, opened 2024-05-07
Waffling around for June

Thinking-waffle:
Fixing bugs, wrong states, typos, all those pesky little things.

Fixed a focus in Tibet giving a wargoal against you instead of a giving you a wargoal against the Chinese warlords.

changed sultunate -> sultanate

tweaked decolonization effects to include the strip of modern day Saudi Arabia now attributed to Britain in 1936.

Changed Portuguese naval characters (more changes to do at the time of this message).

Fixed an event about Danzig giving you a core on Gdynia (Gotenhafen? no?)... also rewrote the text to be more than one simple line.

Fixed the democratic German focus to attack Italy which could grand a claim on Noord Brabant instead of Alto Adigo.


marijn211 (approved):
Diff looks good, Ill do a final error log check then accept

PR #2922, opened 2024-05-08
Bugfixes June

marijn211:
- Removes event check interval change from Spicy that broke several events
- Hide coring decisions when the territory is cored, this is relevant to these 2 specifically because the cores can also be obtained from other sources (a formable and gamerule).


ilkober (approved):
Yeap, that's PR alright

PR #2924, opened 2024-05-10
Channel islands state to avoid issues

TheBrit79:


marijn211 (changes requested):
Please wait with accepting so we don't break all saves again
(Ask someone else if you want to dismiss the review, Im not working on the mod much atm)

hammmurabi:
> Please wait with accepting so we don't break all saves again
> (Ask someone else if you want to dismiss the review, Im not working on the mod much atm)

something needs to be done about them since right now UK can do land trade with all of europe and will never do dday if they aren't occupied

They either need to be made a new state, have the crossing removed, or be removed entirely

One of the 3 and it needs to be in the next update or it's just impossible to convoy raid UK and dday will never happen until august

SwedishNorwegian:
Per Spicy's instructions, this pull request is not to be merged until the next Map Changing pass which is August.

map/provinces.bmp
hammmurabi, line 1: map is expanded to 5632x2112 (for no reason, nothing is added in the additional southern ocean added) which is not divisible by 128 and over the map's size limit causing a giant line to form at the top of the map

hammmurabi (changes requested):
map is expanded southwards from the correct size of 5632x2048 to 5632x2112 (for no reason, nothing is added in the new area) which is not divisible by 256 and over the map's size limit, causing potential crashes and a giant line to appear at the top of the map + heightmap issues

PR #2938, opened 2024-05-22
PRC's mio fix

Charl-Ep:
Fix PRC's focus tree won't active organizations while having Arms Against Tyranny DLC on
An add-on fix mod of this PR is at https://steamcommunity.com/sharedfiles/filedetails/?id=3251842684


marijn211 (approved):
Diff looks good, not sure if we're giving the PRC too many MIOs, but then that can always be addressed

PR #2949, opened 2024-06-02
Pacific Claims

katsuragiyy:
Pacific stuff


marijn211:
Any non-wikipedia sources I can use to quickly verify this?

katsuragiyy:
> Any non-wikipedia sources I can use to quickly verify this?
ok yeah might have screwed up on State 639... 

https://www.doi.gov/oia/islands/disputedislands (1 and 4 are relevant, the rest is not.)

Line Islands and some Elice Islands were disputed because of the Guano Islands Act. 

Hawaii never had any actual claim to the islands it cores in the game, literally.

ALOT of Core-adjustments bcuz they do not really make sense 
(and since most countries in the Pacific are based off Ethnic-Borders, had to squeeze in everyone (Even if it dint make much sense) to one country). 

but even with all of this, dunno if this is worth putting into the game, because claiming somebodies territory reduces their overall diplomatic relations, and if Britian and USA claim each other's stuff non stop, its pretty unlikely that USA will stick with Britian, or the opposite

PR #2958, opened 2024-06-15
Air tech year alignment

theodorhansson:
Slight adjustment to the position of the "year markers" in the aviation tech tree. They are now aligned better. An unused "1950" label was removed. 

Also moved Air Torpedo 3 to match its year.

Tested with and without BBA.

### DIFF
--- a/common/ai_equipment/USA_planes.txt
+++ b/common/ai_equipment/USA_planes.txt
@@ -79,7 +79,7 @@ USA_fighter = {
 		priority = {
 			factor = 0 #100
 			modifier = {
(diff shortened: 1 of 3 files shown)


marijn211:
If it isnt too much work you should comment out the BBA 1950 markers instead of deleting, since it is something we should ideally make something for at some point

marijn211:
Also feel free to restore the whitespace removal commits

marijn211 (approved):

marijn211 (approved):

PR #2962, opened 2024-06-27
Lemuelfixes june24

LemuelSora:
Addition of 3 files:
Baltic.txt to fix a basegame bug that between the baltic nations and the formation of Baltic-Belarus Nation.
During the preparations they can gain full control of all states which breaks the CW and doesn´t work properly.
This has been fixed.
resources_icon.dds     Explained further down
combat_xp.dds 

Fixes:
Fixed the Baltic-Belarus SSR focus as mentioned above.
Fixed Cores for the formation of Finno-Ugria, Formation of Scandinavia decision, Formation of Umayyads, Formation of the Roman Empire, Formation of EU
Fixed England not being able to reign in the Ray as a state needed for that was given to Tibet.
Fixed Soviet focus Solidify the provisional government having the wrong prerequisites.
Fixed a focus for Algeria needing Great war tank research not great war chassis.
Fixed Generic Focustree giving the wrong research bonus for Establish the Armor Corp.
Deleted an unused focus that was hidden behind another one as Cuba.
Fixed minor localisation issues on Cuba and Germany.
Fixed PRC tooltip for Self Preservation showing the wrong values.
Fixed Canada not being able to do a few focuses if they are leading a faction.

Balance changes:
Adjusted which nations are allowed to form EU
Adjusted darien gap province modifier being WAY TOO HARSH.                     -Normann
Adjusted modifier for "crushing Stalin,Hitler,Sweden,Denmark" as Norway      -Normann
Removed a research bonus for Nuclear Technology on Brazil.
Adjusted Self Preservation technology of PRC as it allowed too much XP loss reduction and the trickleback was very weak.
Removed Modern Tank Max Speed from Railroads.
Adjusted Railroad tech to give repair speed.
Minor change to make USA not bline "r56_Combat_Service"

Further Adjustments:
Added a tooltip that shows which advisors you can get when germany does the "Appoint Economics Minister"
This properly works when you change ideology.
Added Yugoslavia getting the proper cosmetic tag, if the soviets offer them to form the united balkan federation

Added Descriptions:
Added Descriptions for all the Ukranian focuses.                    -Zozo
Added Descriptions to some Yemen focuses.                          -Zozo

In progress rework:
Reworking the focus filters for all nations to streamline it and to make sure the focus filters are assigned properly.
New focus filters FOCUS_FILTER_RESOURCES                 Ingame: (Resources)
Instead of using Steel,Aluminium,Oil,Rubber,Chromium,Tungsten,etc... Just use Resources for all of those and for excavation tech.
New focus filter FOCUS_FILTER_COMBAT_EXPERIENCE  Ingame: (Army Experience)
Instead of using Army Experience, Air Experience, Navy Experience

Reworked nations:
Canada, USA, Mexico, Guatemala, Honduras, El Salvador, Costa Rica, Cuba, Haiti, Dominican Republic
This only applies to the focus trees that they have selected by default according to game rules.
I will do the other nations next month and after that look at the non default trees.


marijn211 (commented):
Is this WIP or ready? Open PR is for stuff ready to be accepted but description has an in-progress part?

Regardless we believe XP filter should probably not be merged like resources should

marijn211 (changes requested):
The "fix" to ABC_shared has two issues:
A - NSB tanks need a design when being added, or you end up with tanks with no gun
B - you are breaking the effect and bypass for people without NSB

LemuelSora:
A - I undid the XP merge.
B - I showed that the tanks work just fine for anyone who owns NSB as france has 2 interwar light tanks to pick from and also the bypass works correctly. Do you want me to make a code piece in case the user doesn´t have NSB?

marijn211:
Thing B was meant to say without, oops

marijn211:
Please do not shift around the focus filter placement within focuses next time

marijn211 (approved):

PR #2964, opened 2024-07-01
A little to the right, a little to the left, to whom does Berlin belong?

katsuragiyy:
small and mostly overlooked things

rushed, so if there is any mistakes then my bad


marijn211 (changes requested):
Please do not create new country leaders with create_country_leader, use character files

marijn211 (approved):

PR #2968, opened 2024-07-06
Nepalese things

Thinking-waffle:
trying to add historical people for Nepal, one step at a time. It will need to be checked so that campaign of various ideologies are possible. The "sudden disappearance" of various reformers in 1941 makes the effort a bit more difficult. The absence of a fascist party also makes it harder. The Rana clan was just extracting the country and enjoying the status quo. There are still at the time I am writing this 3 characters that should ideally be removed, they have british names and no traces of presence in the country. There are also filler characters as chief of the navy and air theorists.


marijn211 (approved):
Will check the Greece changes in-game, then accept if theres no errors

PR #2971, opened 2024-07-09
August bugfix

marijn211:
Bugfixes and an event option in "Fate of Korea" to divide it among USA/SOV, the receiver playing one of those (it establishes North Korea and South Korea as supervised governments to represent the period they were occupied)


marijn211 (commented):
@TheBrit7 I see you removed some NIC portraits, but these are still referenced

TheBrit7 (approved):

PR #2972, opened 2024-07-11
Smol bugfix for Honduras and even smaller change for Turkey

dogukans1:
As for Honduras:
1. "Support the USA" focus code is improved to make sure Honduras is joining USA against Fascist Germany, and accounted the case where USA takes over Allies. This focus' description explicitly states that Honduras wishes to join USA that fights German Reich
2. Added code to remove "Monroe Doctrine" spirit for every focus that used to remove US guarantee as US no longer simulates that by guarantees but rather a spirit

As for Turkey: "Modest Economic Liberalisation" focus is locked if Etaism focus is completed. This focus makes Celal Bayar the PM, which locks all paths associated to Etaism and costs at least 210 PP to get rid of him with a suitable PM


marijn211:
Although the focus description may target Germany, we prefer the focus to be flexible

dogukans1:
Reverted to the old state. Added bypass checks in case UK and/or US is no longer Democratic so a Democratic Honduras enjoyer doesn't get themselves locked on an ahistorcial game

marijn211 (changes requested):
Fixing some typos and seeing if the focus is flexible enough, then accepting

marijn211 (approved):

PR #2974, opened 2024-07-17
Adding an accepted achievement to R56

dogukans1:
This achievement was accepted on Discord, however either forgotten to be implemented or replaced by Hispanoamerica achievement. Refer to https://canary.discord.com/channels/337306241149960193/1045418941822029824/1105389127098179674 for details.

Things to note
1. Fascism requirement is strictly because the copy-pasta, otherwise it's fine to just have it removed
2. Description itself is also inspired by closing notes of the copy-pasta, feel free to change it with something generic.


ilkober (changes requested):
Right now it fires if you are just not in the faction.
Meaning you can join one, achieve the conquest goals and then just leave it to fire the achievement

dogukans1:
I've made some changes now, hopefully this fixes the issue you've mentioned? I've thought achievement would just invaildate itself once "Not in faction" check has failed but apparently not

marijn211:
Personally I think its a bit pointless to do this achievement for a meme, when Peru can the do the same with actual content

Other than that only the localization seems flawed

PR #2978, opened 2024-07-28
Liechtenstein focus tree

TheBrit7:


TheBrit7:
Lingering errors.

marijn211 (approved):
Haven't playtested but looks good in-game, so does the diff and no errors

PR #2980, opened 2024-07-30
Waffle august debug

Thinking-waffle:
Ensuring the branch is up to date would help I guess.

Fixes things about Fryderyk Wettin being put on the throne of Poland in a rather inelegant way.


hammmurabi (approved):
looks ok, no errors

PR #2983, opened 2024-08-02
August Beta Bugfixes

ilkober:
Accept late


marijn211 (dismissed):
Should be checked whether the lack of OOBs is an issue

marijn211:
@TheBrit7 one of the adjecencies has an error

hammmurabi (approved):
falkland strait crossing has an error but seems to be working as intended, and the removal of the OOBs doesnt seem to have any issues either, merging.

PR #2988, opened 2024-08-07
September Bugfixes Branch

ilkober:
Fixes
Removed reference to the incorrect focus in Peruvian election event. [LaR] Portuguese ship mutiny will correctly determine the amount of ships lost to mutineers. [NSB] FIxed a tooltip about Meletius rising in ranks.


ilkober (commented):
Fixes
The game can now handle the effects of Platine Unification and will no longer crush.
[BftB] Only fascist and monarchist Bulgaria will be able to form the Third Bulgarian Empire.
Yalta Conference now requires Soviets to be at war with Germany. No second Hitler-Stalin pact is allowed here!

marijn211 (approved):

PR #2991, opened 2024-08-16
Made a pull request on accident but hey it's just about nepali names so that one can't break things

Thinking-waffle:


marijn211 (approved):
checked and GFX doesnt throw errors, rest of the diff seems good

PR #3006, opened 2024-09-27
minor jordan update

hammmurabi:
Added a small tree for Jordan and changed the way british protectorates enter the Arab-Israeli war.


common/ai_strategy/r56_ENG.txt
```diff
@@ -1602,4 +1602,53 @@ r56_ENG_dont_turn_ECU_PRU_war_into_WW3 = {
 		id = "ECU"
 		value = -75
 	}
+}
+
+r56_ENG_dont_join_jordan = {
```
marijn2112, line 1607: If UK has any allies, those would probably call them in indirectly, although if only dominions, this should suffice

marijn211 (dismissed):
jordan.6.a seems to be unclear, looks good otherwise

SpicyAlfredo (approved):

PR #3010, opened 2024-10-05
Last minute tweaks

katsuragiyy:
little things


TheBrit7 (dismissed):
Please use correct capitalization and full stops on event options.

Also why change Ghana to Ashanti? This is a region, not the country as a whole.

marijn211 (approved):

PR #3018, opened 2024-11-14
Speculoos things

ghost:


marijn211 (dismissed):
Benes was intentionally not the country leader, the portrait is also from 1946 instead of vanilla's 1942 one

SpicyAlfredo (approved):

PR #3027, opened 2024-11-20
Minor fixes for localization, SH tanks and airplane modules

ma44:
- Ported over the localization for jet axial engines and ballistic missiles
- Super heavy tanks were recently rebalanced in vanilla, ported them over here because it feels real bad to have worse attack modifiers into terrain than a landcruiser whilst having a fraction of the stats.
- Jet engine axial modules are no longer deleted from the mod


SpicyAlfredo (approved):
Thanks for the assist

PR #3056, opened 2024-12-23
Update ARG_l_english.yml

Harain:
fixed 4 missing locs (were there but misspelled)
also reordered them to match in-game to be easier to track there a few redundant things i left alone

as a side-note, most vanilla MIOs also appear in-game as Argentine, might want to look into it


marijn211 (approved):
Currently good, wont merge yet in case you wanna throw some more bugfixes in

Harain:
Not really sure how to work with more complex Pull Requests, never used this before yesterday, always worked by commiting straight to repo.

Need to add a new file, copied from vanilla... and as I'm typing this I see right below "Paste, drop, or click to add files" lol

...But it won't let me add an asset file, says it's not supported

Need to copy the vanilla file _LaR_units_infantry.asset to comment out the POR_infantry_entity there to fix the following error:
`[01:14:40][no_game_date][pdx_entity.cpp:2171]: Duplicate of POR_infantry_entity added to entity system`

SpicyAlfredo said I'd get access eventually, when someone comes online, so I'll wait till then

PR #3078, opened 2025-01-12
Spanish civil war no longer removes cores

marijn211:
Also fixes some Spanish states being impossible to re-core since the automatic removal of cores causes no resistance to appear


marijn211:
Currently entirely ready _except_ for some anarchist cores which refuse to disappear for some reason

Thinking-waffle (approved):
works, tooltip needs adjustment

PR #3079, opened 2025-01-15
Loc key collisions

Harain:
just a draft for now, a bit over halfway done


Harain:
fixed over 2k duplicates, only 8 left which for some reason are still crating a duplicate error even tho in replace
moved them all to localisation\english\replace\r56_replace2_l_english.yml to keep separate

tried to keep criteria similar, but it's hard with so many:
- deleted exact vanilla copies (unless close to a desc where I left it commented
- lots of duplicates in same file deleted or commented if unsure of which text was better
- when deleting duplicates, always kept the ones either different from vanilla or most recent
- fixed dozens of locs that were missing _desc or something similar, same for some event locs
- i'm probably forgetting other special stuff, but it's all in the commit descriptions

marijn211 (changes requested):
Since this is based on the bugfix branch, and contains quite a lot of changes, I will check this after accepting the bugfixing branch

SpicyAlfredo:
> Since this is based on the bugfix branch, and contains quite a lot of changes, I will check this after accepting the bugfixing branch

Eyeballing it, everything looks right. If I don't have time to check before the first this might just be a 'fuck we ball' merge and we fix quirks with it before the 7th

marijn211 (changes requested):
By moving parties_l_english.yml out of the replace folder (which in principle is a good idea), we stop inheriting lines that are only in the vanilla file of that name, some of which seem to be unported

marijn211 (approved):
Double checking rn but looks good

PR #3080, opened 2025-01-16
Finnish BOP Adventure

Simpyslav:
- Loc Change for Personal Agenda (Wasn't showing up at start up at all - changing the wording to fit the democratic ideology.
- On Action changed due to Leader Ideology change
-  Should be fine and work as intended
Ported over from RP


marijn211:
I didnt know of this issue, but the fix is wrong

PR #3086, opened 2025-01-27
Italy Fixes February 2025

BiccyThiccy:
I would just put this on the February bugfix branch, but this should probably get a lookover before it gets merged, as I do not work with ITA files and I am not certain of some intended choices.

Fixes:

1: Special project changes from vanilla were not properly ported over. This led to the focus ITA_caligulas_pride always being unavailable since GVD.

2: Fixes one capitalization bug

3: Comments out ITA_pietro_koch from the Italian focus tree. Waffle commented him out of the history file, so him being here is spamming our error log. I am going to assume the intended behavior is that the advisor is supposed to be gone from focus tree references.

### DIFF
--- a/common/national_focus/italy.txt
+++ b/common/national_focus/italy.txt
@@ -3857,7 +3857,7 @@ focus_tree = {
 		relative_position_id = ITA_fiocchi_munizioni
 		cost = 5
 		available = {
-			has_tech = radio_detection
+			is_special_project_completed = sp:sp_air_radar
 			has_full_control_of_state = 159
 		}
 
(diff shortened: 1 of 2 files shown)


BiccyThiccy:
For the patchnotes:

- The Italian focus _Caligula's Pride_ will now be available upon completion of the Super-Heavy Battleship Special Project. 
- Various effects utilizing Special Projects in the Italian focus tree have been implemented

SpicyAlfredo:
These changes are fine. Use Vanilla over our changes when it comes to SP’s focus effect changes. If you still under the hood on this could you port over the midget sub focus effect changes aswell. If not I can port them over tomorrow.

BiccyThiccy:
> These changes are fine. Use Vanilla over our changes when it comes to SP’s focus effect changes. If you still under the hood on this could you port over the midget sub focus effect changes aswell. If not I can port them over tomorrow.

Affirmative senior 🫡

PR #3090, opened 2025-01-30
Belarusian noble

ghost:


katsuragiyy:
should be converted into a draft until a conclusion is reached. 
Currently huge debate about weather or not to to replace the current Non-Aligned leader

marijn211:
> should be converted into a draft until a conclusion is reached. Currently huge debate about weather or not to to replace the current Non-Aligned leader

If you could summarize the discussion into a message here before tomorrow evening, that would be poggers

katsuragiyy:
@marijn2112 Many arguments on both sides. i cannot summarize into a single message but you can look at `#katsura-island-hopper` on Discord

PR #3092, opened 2025-01-31
East indies Reworked #3 - Portrait Updates, Indonesian character updates

KrasnaMerah:
Intended changes:
- You can unlock the Islamists (PSII), Right-Nationalists (Parindra), and Dutch Loyalist advisors if you backed PSII in the colonial reform path and already free
- You can unlock Parindra, Left-Nationalists (PNI), and PSII advisors if you backed Parindra in the colonial reform path and already free
- More monarchs for Parindra monarchist path
- Portrait updates for: Thamrin, HB VIII (Army), Sam Ratulangi, Manusama, Soumokil, Tou Samouth, Ho Chi Minh with appropriate hidden "character aging" events


marijn211 (commented):
Why does INS_munawar_musso require MtG in his advisor role? Doesnt seem like the trait is MtG-only

KrasnaMerah:
> Why does INS_munawar_musso require MtG in his advisor role? Doesnt seem like the trait is MtG-only

Probably a forgotten leftover from East Indies Reworked. Yeah, just remove the DLC requirement.

> Better way to fire lategame event

Thanks, if i have the time, i'll do it for any Indonesian characters who has both old and young portraits

> Simplified some advisor checks

Oooohhhh, so that's how you do it if the advisor becomes a leader, thanks.

marijn211 (approved):

PR #3100, opened 2025-02-02
Portugal post-Beta Changes

BiccyThiccy:
- Nerfing the late-game stability rebound slightly.
- The focus **Draft Portugal's Own Labor Charter** now has one line skip.
- Removed the compliance/resistance requirements from the decision **Relocate the Capital to Madrid**.
- The party name change from **Radicalize Salazar's Allies** has been moved to **March on Lisbon** for better player clarity.
- The focus **Experiment With Radar** will now give the generic RADAR idea, as the focus wasn't really worth doing before.
- The focus **Complete the Separation** will now create an Aerial Research facility, to give a small buff to those who want an air-centered Portugal build.
Most of the Restoring Valiance to the Nation decisions have had their effects reworked into longer-lasting, more expensive temporary modifiers.
- Portugal's Fascist Demagogue has been turned into a 75 political power advisor that unlocks upon the completion of **All Against the Bolshevik Threat** to slightly balance how unneeded he is in the R56 tree.
- The focus **Modernize Our Infantry** has had its effects reworked: now gives 2 * 175% speed instead of 3  * 150%, and now gives a 200% bonus to Improved Desert & Jungle Clothing. I made this change because an important trigger in my tree calls for it.
- The decision **A New Status for the Colonies** will now transfer Sao-Tome and Cape Verde if they are controlled by your subject (the decision cored them already). It will also now be available for the Communist path upon the completion of **The Iberian Social Revolution**, if you choose not to withdraw from your colonies.


TheBrit7 (approved):

TheBrit7:
Merge when ready.

PR #3101, opened 2025-02-02
VP name language toggle button

marijn211:


marijn211:
Should be ready, but I still want to turn the FIN_change_name_of_state_or_vp scripted effect into the new form of state renaming

TheBrit7 (approved):

PR #3102, opened 2025-02-02
Medium naval bomber

Harain:
this is just to see if it can actually be done

atm it allows design, but doesn't actually save it

would like to know if we actually want this in mod, before spending more time trying to implement it

it will probably have large repercussions in many areas, so we need to decide if we want it


TheBrit7:
What is the purpose of this PR, exactly?

BiccyThiccy:
Inactive? 

Anyway, hope to see you back on the mod some day my friend 💝

PR #3124, opened 2025-02-27
TUR stuff

katsuragiyy:
quick fix


marijn211 (changes requested):
Mismatch between Rifat Özdes' large and small portrait, that large portrait has a linked small one

marijn211 (approved):

PR #3131, opened 2025-03-08
Circassia Formable

InfraTheIdiotic:
Circassia formable for Kabardino-Balkaria.


marijn211:
Could you explain the basis for this?

InfraTheIdiotic:
Circassians were exiled in the 19th century after the genocide, given that Kabardino-Balkaria's fascist tag is the 'Circassian Realm' I figured restoring the old borders of Circassia would be a logical step forward. While I can't think of any figures for it outside of the Mountainous Republic's Circassian politicians (perhaps I could have added one in hindsight), I wanted to give some detail to one of these many 'restored country' alternate ideology names.

InfraTheIdiotic:
Messed it up, wanted to do something new

PR #3133, opened 2025-03-10
UK communist path changes / tweaks

katsuragiyy:
for more accuracy


BiccyThiccy:
Can you give a small list of what this PR changes?

It would not only make it a lot easier to tell what you're doing, but it will be needed for the patch notes

katsuragiyy:
@BiccyThiccy Previously the UK communist path was slightly broken due to my altercations in the Pacific (Seizure of Canton Island by the USA, swapping around state ownership, etc), and now that i have MtG, i decided to come fix my un-intended bugs here.

It also makes it so that some countries switch to monarchism, despotism, and etc when released (to better repsent cases like Tonga where they still maintained their native monarchy while being under British sovereignty).

It fixes a tag-related bug between Barbados and Barosteland.

It may also contain more tweaks and updates, but i'm currently busy with something so i cannot give it my full attention. Hope this helps!

BiccyThiccy (approved):
Doesn't add to the error log and I didn't eyeball any issues when firing the focus's. Most of the changes seem logical albeit insanely unnecessary.

Iraq should probably get Kuwait via Withdraw from Disputed Territories, and Britain still hangs onto one Somali island they don't core, but I cba and you can change things later if you want.

The decolonization branch really needs a rework as a whole, both as a no-dlc alternative and because this branch is just horrible. If you want to contribute to a less pointless tag than Samoa maybe you should consider it.

Approved

PR #3134, opened 2025-03-13
Fixes Brazil not getting its cores when it forms the united states of south america

Moccha-Bee:
Like the tittle says.

### DIFF
--- a/common/national_focus/brazil.txt
+++ b/common/national_focus/brazil.txt
@@ -4980,6 +4980,11 @@ focus_tree = {
 		completion_reward = {
 			set_cosmetic_tag = BRA_united_states_of_south_america_cos_tag
 			custom_effect_tooltip = BRA_united_states_of_south_america_tt
+			every_state = {
+			limit = {
+			is_on_continent = south_america
+                  }
+                add_core_of = ROOT
 		}	
 	}
 


Moccha-Bee:
Btw if the commit is fucked tell me, i havent used github in 2 years

BiccyThiccy:
This isn't how the vanilla code for United States of South America functions.

I can just port the exact way vanilla coded the join events, which adds the cores upon acceptance, but your change makes it even stronger than what vanilla already does, so I don't see the point of adding it when it's already a laughably strong section of the tree.

I don't know if the coring upon acceptance functionality was intentionally removed. If it wasn't, I'll have it fixed next update, but anyways I am closing this request because there is a better fix for it.

Moccha-Bee:
'

> This isn't how the vanilla code for United States of South America functions.
> 
> I can just port the exact way vanilla coded the join events, which adds the cores upon acceptance, but your change makes it even stronger than what vanilla already does, so I don't see the point of adding it when it's already a laughably strong section of the tree.
> 
> I don't know if the coring upon acceptance functionality was intentionally removed. If it wasn't, I'll have it fixed next update, but anyways I am closing this request because there is a better fix for it.

Alright, thanks for fixing it

BiccyThiccy:
👍 Np

PR #3136, opened 2025-03-13
Zozo's localization stable

Volttsie:
Slight LIE localization, slight changes to warlord focus tree to make it in line with Vanilla update.


BiccyThiccy (approved):
Error log and text log look good. It's merging time 😎

The Liechtenstein localization additions aren't perfect but it's a lot better than nothing. The Warlord changes are identical with vanilla's, plus I see no reason to deny slightly buffing them. The Yugoslavia changes make the worst path in the tree a bit better, and some of the additions I did reduces unneeded checks and polishes things further.

PR #3139, opened 2025-03-24
Xaelor's Czechoslovakia Contributions

Xaelorichero-02:
I want to make RT56 slightly better. Basically I added some extra focuses for the communist path for Czechoslovakia with some tweaks here and there. Also the Czechoslovakia-Poland formable has been added.

Full list of changes with comments:
- New focuses for Czechoslovakia: Rally Against the Nazis, Treaty of Friendship with the Soviets, Divide Poland, Unite with Poland
- Added focus shortcuts for Czechoslovakia
- "New" advisors and generals to Czechoslovakia: Jan Antonín Baťa, Bedřich Neumann, Josef Zmek, Richard Tesařík (They were in the code and fully functional so why not use them
- New national spirit for Czechoslovakia: Divided Nation (If Czechoslovakia becomes the Czech Republic it will be removed)
- Nerfed King of the Shoes trait for Baťa by 1/3 (-5% inf, civ construction speed)

image
image


ilkober (commented):
The country tag is still called CZE_westslavia

Would also fit better to build a union state in pro-Polish and not pro-Soviet branch, since USSR didn't like Poland and didn't want to have them strong while POL-CZE were fine*ish

Xaelorichero-02:
> The country tag is still called CZE_westslavia
> 
> Would also fit better to build a union state in pro-Polish and not pro-Soviet branch, since USSR didn't like Poland and didn't want to have them strong while POL-CZE were fine*ish

Remember in this case the Czechs are fully independent and mostly equal to the Soviets. Something like Communist China before the split.

events/CZE_Czechoslovakia.txt
```diff
@@ -4281,6 +4291,23 @@ country_event = {
...
+			swap_ideas = {
+				remove_idea = CZE_divided_nation
+				add_idea = CZE_divided_nation_2
+			}
+		}
+		if = {
+			limit = { has_idea = CZE_divided_nation_2 }
+			custom_effect_tooltip = CHL_skip_one_line_tt
+			add_political_power = -50
+			add_stability = -0.5	#To account for national spirit
```
marijn211: -50% stab??

marijn2112 (changes requested):
Either disable Bata or the generic trait industry advisor in my opinion, otherwise you can stack like 25% construction speed on civs?

Xaelorichero-02:
> The country tag is still called CZE_westslavia
> 
> Would also fit better to build a union state in pro-Polish and not pro-Soviet branch, since USSR didn't like Poland and didn't want to have them strong while POL-CZE were fine*ish

The country tag doesn't matter at all if it's not a duplicate.

marijn2112 (commented):
Error log and diff looks good, just gotta check in-game later

PR #3142, opened 2025-04-03
Treaty of Trianon fix

someone19204:
Fix Hungarian "Renounce Treaty of Trianon" Gotterdammerung focus.


common/national_focus/hungary_wuw.txt
```diff
@@ -8513,10 +8513,10 @@ focus_tree = {
 						days = 1
 					}
 				}
-				ELSE = {
```
marijn211: This is identical

PR #3144, opened 2025-04-04
Fix Renounce Treaty of Trianon

someone19204:
Fix Hungarian "Renounce Treaty of Trianon" Gotterdammerung focus not removing the idea


someone19204:
I had to create a new pull request due to an issue with git. I'm new, so I apologize if that's wrong.

marijn211 (approved):
Matches vanilla so gonna mostly blind accept

PR #3145, opened 2025-04-05
New Portugal National Spirit + Polish and Balancing of the Tree

BiccyThiccy:
- Portugal now has an illiteracy national spirit, to simulate the fact that the nation only had about a 40% literacy rate.
     - The national spirit starts with -15% research speed, and a +15% cost modifier on all doctrines.
     - Focus's in the political/economic section of the tree that provide a research slot also reduce this spirit by ticks of 5% each, the          second time you reduce the spirit turns it into a temporary penalty that disappears after 2 years.

- I buffed the colonial focus's to make them more (as they were historically) relevant. They are also now worth keeping, when before they were nothing but a small source of manpower.
     - Develop the Angolan Interior has been bumped up to 70 days (was 56), and now provides more oil (it became significant in the 1950s, but this isn't a focus the AI does until very late, so I think it's acceptable), and it provides some Steel (Angola had fairly significant Iron deposits, about 3-4% of France's from what I could find).
     - Develop Mozambique no longer gives Aluminum, as it doesn't seem historical for it to give very much; it now gives a tiny bit of steel to account for coal in the state, and a new national spirit that provides a small bonus to CG, wartime stability, and inf-equipment/support-equipment production. This spirit is to account for the significant cotton and other cash crop exports that are useful in the production of wartime material.

-  I did a balance pass of the whole tree, TLDR since I changed too much:
    - Tank, mechanized, air and naval builds will now be somewhat stronger.
    - The flow of the tree should feel more natural
    - The democratic and historical paths were slightly buffed, while the fascist and communist were slightly nerfed
    - Most research bonus's were increased to account for the new national spirit, but some more than others
     
- The only other thing of note that I did was that I redid most of the decision icons so that there were less duplicates, and more high quality ones in use. I also re-arranged some so that alliance decisions were always together.


BiccyThiccy:
Sorry for last minute stuff, but they're so tiny they shouldn't affect the checking process. I'm done now.

marijn211 (approved):

PR #3146, opened 2025-04-05
Tweaks for Albania

Xaelorichero-02:
Created extra icon for the Albanian focus tree and changed a few others to better match the focus.
Added `ALB_` in front of every focus id to match modding standards (localization and stuff related to it in the Italian focus tree adjusted)


BiccyThiccy (approved):
The new icon looks good, the shine is functional, and there's no new errors in the text or error.log files based on the launch, and in-game when hovering over every changed focus. I appreciate the polishing job, I see no reason to delay approving this. 

The only thing for future reference @Xaelorichero-02, If you are going to edit a localization key for any reason, make sure the name you change it to matches the text in-game. I know these weren't yours originally but you should've done it since you were overwriting it anyways.

PR #3154, opened 2025-04-23
Maps-May

katsuragiyy:
Province and State tweaks. does not break saves iirc
Tweaks to State Batambang (925) in Cambodia.
Tweaks to State Champasak (924) in Laos.
Tweaks to the entire Malaysian Peninsula. Johor (951) is now more accurate to its irl borders. Kelantan has been merged with Terengganu (960). and Kedah (914) has been adjusted. Singapore (336) has also been made smaller.
Brunei (1014) was also tweaked to look more like its irl counterpart.

And the most noticeable change is Canton Island (642), which has been resized and reshaped.


TheBrit7 (approved):

TheBrit7:
Merge when ready.

PR #3155, opened 2025-04-24
Alterations for Xinjiang and Tunganistan/Khotan

InfraTheIdiotic:
**_Changes:_**

**- Changed Tunganistan's name to Khotan Clique, other ideology names refer to East Turkestan and Khotan. Adjective for Khotan is now Khotanese.**
Rationale: 'Tunganistan' is a Western exonym that might have been unheard of in the area, Wikipedia refers to it this way in its article. Ma Hushan's regime in Khotan issued banknotes under the Khotan district which seems to me like the closest there is to an official name for the polity. RT56 uses official names for all other states, and I think Khotan is much more suitable for TNG in that regard. It is named Khotan Clique for uniformity with the other Chinese warlord cliques.
Chinese-language blog post about the 'Ma Hushan banknotes' that I was looking into:
https://9collection.blogspot.com/2012/01/blog-post_12.html
The adjective for Khotan, Khotanese (now Hotanese), can be found in use by a verbatim google search.

**Changed Xinjiang's alternate ideology names, removed references to East Turkestan, non-aligned is now also the Xinjiang Clique. Adjective is now Xinjiangese.**
Rationale: Avoids Han-led Turkic states. Ideally, I think TNG shouldn't reference it too and East Turkestan should be its own tag which could spawn in the historical Ili rebellion, but I don't think it's worth it without proper content for Xinjiang. The adjective, Xinjiangese, can be found in use and appears in online dictionaries.

**Changed Xinjiang's non-communist flags, a generic one with 'Xinjiang' in Chinese characters for democratic and fascist and one with 新 on the ROC flag for non-aligned, giving uniformity with the other warlords**

**Changed Khotan's flags except the communist one, ROC-based one with Hotan's name for nonaligned, Kök Bayraq for fascist and democratic**
Rationale: The internet has understood the previously used flag with the star and crescent to be fictional, Wikipedia has dropped it as there was no source. I have replaced it to one that fits with the other warlords as Ma Hushan was loyal to the Republic.

**Altered Xinjiang's parties, removing Uyghur-related ones and correcting the Xinjiang People's Anti-Imperialist Association's name**

**Altered Khotan Clique's parties, Khotan District Administration for nonaligned, historical Committee for National Revolution for East Turkestan, changed All-Union Communist Party to Communist Party of China**

Note: If I knew how to make vanilla-style portraits with awful sources, I would have corrected Ma Hushan's portrait as his current one is someone else.


DutchPlanDerLindeSus:
The "official" name of Tunganistan was "Autonomous Military Administration of Southern Xinjiang", however, next to it is the name "Nankiang", and after some talking with ChatGPT, it told ne that "Nankiang" means "Southern Frontier/Territory", while Xinjiang means "New Frontier/Territory", so imo u should rename it to "Nankiang Clique". And also, it appears they did use the current in-game flag during the time also.

Source: http://w.ethnia.org/polity.php?ASK_CODE=XJ__&ASK_YY=1936&ASK_MM=01&ASK_DD=01&SL=en

InfraTheIdiotic:
Ethnia.org is not reliable, none of that is sourced. Yes, Nankiang means southern frontier but the state did not control the entire region, just the area around Khotan south of the Taklamakan desert. Please don't treat it as an end-all source when it has many strange names elsewhere and fictitious flags. 'Autonomous Military Region of Southern Sinkiang' sounds like it is descriptive, it wasn't a formal polity and if that was its official name, you would hear it outside of ethnia.org on the internet.

DutchPlanDerLindeSus:
I would still prefer that it be called the "Nankiang Clique" as it actually is soirced at least here, and makes more sense than "Khotan Clique". Plus the flag should probably be kept as well.

InfraTheIdiotic:
The actual borders did not resemble Nankiang's, only R56's inaccurate one does. It had no control over the larger Nankiang cities such as Kashgar (though Ma Hushan had desire to conquer them, and the ousting of Sheng Shicai for that matter). His territory revolved around the city of Khotan rather than the region.

And why should we keep the flag when it has no backing? We don't use made up internet flags for the others, just variants of the actual flag they flew - the ROC flag - with Chinese characters to differentiate them in-game. Sure, the flag looks nicer but since when should we put aesthetic over accuracy? And it would make it stick out in that regard for no good reason.

DutchPlanDerLindeSus:
I am unaware of "Nankiang" being a region with set geographical features. Please elaborate on that, as when i would search it up nothing would come up. Maybe try searching on chinese web also. However, the flag can be changed if not all sources agree on the red flag being the flag. But the name, i would still prefer to be "Nankiang Clique".

Also, please dont treat me as someone who isnt well researched in this topic. I have dont a lot of research on Xinjiang and Chinese Warlords in the past, however, a lot of things have not been implemented, for example the true borders of Xinjiang-Nankiang with Kashgar still in their conttol etc. as that would require the addition of another state, which breaks saves etc. etc.

Ethnia.org is semi-reliable, sometimes you will find some altered flags there, but thats bound to happen. Name-wise they are almost 100% accurate from what ive seen.

TL:DR Make the name "Nankiang Clique", change the flag to ROC flag with "Nankiang" in Chinese. (If sources dont agree on the red flag)

DutchPlanDerLindeSus:
Will correct myself, as Nankiang is the older romanization (Wade-Giles) and i am pretty sure the standard for Chinese names we use is Pinyin, so the name should be "Nanjiang Clique". Alternatively, i also recall it being referred to as "Huijuang" meaning smth like "Land of the Huis", Huis being the Muslim Chinese that controlled the land practically, but i am sure u know this already.

InfraTheIdiotic:
Sorry if I came across that way.
I think it's agreed upon that the southwest area with Kashgar and Yarkand is Nankiang. Not easy to research quickly in Chinese google, but going by geography and what can be seen I think it'd be alright to consider it so.
The problem with Ethnia is that we are just taking their word for it - it's still unsourced, but Khotan/Nanjiang/Tunganistan clique has the problem of being so informal of an administration. However, considering that those Khotan district banknotes were made under their auspices, Khotan district seems like the most serious local government in the area, although they weren't exactly the same thing. But TNG revolved around it, so Khotan Clique seems many steps closer to accuracy than Tunganistan to me. I wouldn't complain as much about 'Nanjiang Clique' but I haven't seen any sign of any authoritys activity there going by that name.

InfraTheIdiotic:
PS: will change the flag as soon as I can depending on the leads' verdict

DutchPlanDerLindeSus:
At this point its a thing of preference really, as both names dont have an actual source. In my opinion, its not wise to tie the country to a specific city as the name, bcz the capital can always change. Thinking about it more, it might be wise to just call it "Huijiang", which, as ive said, means "Hui Frontier". This is in comparison to the actually used name for Mengkukuo, which is the Chinese name "Mengjiang", meaning "Mongol Frontier". It would make much more sense than both of the proposals which tie the country to the geographic region. This way it would accentuate the ethnic makeup, or at least the preferred ethnicity within the region.

DutchPlanDerLindeSus:
I have forgotten what Mengjiang is called in game tho, with its full name, so you could check that as well.

InfraTheIdiotic:
I thought Hui at the time meant Uyghur, which were the vast majority of the entire Xinjiang at the time. The Muslim Chinese in Xinjiang were called Tungans, which is what the nickname of Tunganistan refers to. Correct me if any of this is wrong.

IIRC Mengjiang, as it was given content, goes through its historical names over time (Mongolian Autonomous Committee etc > Mongol Military Government > Mongol United Autonomous Government > Mengjiang). Mengjiang was an official name they used.

I guess this can be left to the leads, both our cases are made and they can decide based on that and their own research. I also thought that if TNG defeats Sheng Shicai it can claim the name of Xinjiang Clique through a decision and cosmetic tag, but I did not implement that. If the leads would like it I can implement that into this too. I don't think that it's intended in-game for for TNG to just take part of SIK's land minus Urumqi/Dihua as they have cores on the province, so they wouldn't have any capital other than Khotan or Urumqi. I haven't worked with scripted peace deals but I imagine that could be done too if we want that to be a certainty.

DutchPlanDerLindeSus:
Yes, they were called Tungans (Dungans), but thats by the Russians. The Chinese name for them is "Hui", and they are Chinese but Muslim: https://en.m.wikipedia.org/wiki/Hui_people

Hui and Uyghur are not the same tho, as Uyghurs are of turkic origin, while the Hui are esentially Han Chinese, but Muslim. They had their falling-outs in the 30s with the rebellions, as many were led against the Mas also.

The source for Huijiang is the Chinese wiki: https://zh.m.wikipedia.org/zh-hans/%E5%9B%9E%E9%83%A8

It refers to it as the "area south of the Tien Shan mountains, also known as Nanjiang, as the residents of this area were called the Hui people"

InfraTheIdiotic:
Not really relevant but it seems to me like that WP article says Hui refers to the 'ancestor of the modern Uyghurs'.

InfraTheIdiotic:
Alternatively we could call it the New 36th Division but that would refer to the army

InfraTheIdiotic:
And they were also known to use the Blue Sky and White Sun flag so I could change it to one like XSM's in the mod, but I'll wait for the verdict

TheBrit7 (approved):

TheBrit7:
I have decided to approve this pull as the rationale for the change in name is sufficient. I would ask that someone merges it once it is confirmed that it has no errors and/or conflicts. Thanks.

InfraTheIdiotic:
Made some last minute changes as I realised the party wasn't the most accurate choice and you seem to approve the name Khotan Clique so I changed the flag. There don't seem to be any errors that I could see, it is very simple as it's just loc and gfx changes. Doing CTRL+F on SIK and TNG in the error log didn't bring up anything.

marijn211:
Error log is clean, diff looks clean, is up-to-date, will do an in-game check and merge if thats all good

marijn211 (approved):
This is gonna be mildly annoying when I add East Turkestan but whatever

PR #3156, opened 2025-04-25
Update 00_ideologies.txt

Gold3n-Bat:
Added Ideologies for Monaco includes:
Anarcho-Individualism - Benjamin Tucker
Kleptocratic - Basil Zaharoff
Monarcho-Syndicalism - Charles Soccal (optional - mainly for flavour)

### DIFF
--- a/common/ideologies/00_ideologies.txt
+++ b/common/ideologies/00_ideologies.txt
@@ -142,6 +142,10 @@ ideologies = {
 				can_be_randomly_selected = no
 			}
 
+                        monarcho-syndicalism = { #Only ideology for Charles Soccal - as his party, praised him as a monarch/god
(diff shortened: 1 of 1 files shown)


Thinking-waffle:
Kleptocracy is not an ideology. It's a behaviour of the rulers.

They can be oligarchic and kleptocrats if they are corrupt but nobody would proudly be kleptocrats.

Gold3n-Bat:
Fair enough, Basil Zaharoff would proudly be kleptocrat, as it's literally would he would have done. Oligarch works just as well

marijn211:
Put this on a PR with people actually using this, as well as the associated localization, that monarcho-syn ideology name also seems inconsistent with the "-"

PR #3157, opened 2025-04-25
Update MNC - Monaco.txt

Gold3n-Bat:
Massive Update - Includes all leaders, advisors (millitary & political) 
added Helicopters because of Maurice Farman


marijn211 (changes requested):
You're missing the character file

Gold3n-Bat:
The character files are being made and going to be added, I have updated the MNC since. if you can get decline the request, I can update it, next patch

PR #3160, opened 2025-05-01
Monaco -  Added Leaders, Advisors, Traits. Updated Ideologies

Gold3n-Bat:
Swapped Benjamin Tucker with Charles Soccal as he is a Syndicalist.

Added Syndicalist Advisors from Charles Soccal Party.

Minister of States as Advisors

Founder of the Grand Prix & Rene Blum as Advisors 

If there is anything I missed give me a shout


TheBrit7 (changes requested):
Some issues:
* Anarcho Individualism is similar to anarcho captalist. Kleptocratic is similar to oligarchism. Use them instead, no need for these to be ideologies. 
* Helicopters from game start? This is like a 37-38 tech. No other nation has this.
* Benjamin Tucker was an American anarchist who just happened to die in Monaco with his family. He had no actual connection to Monaco.
* A lot of these characters, who you are adding as leaders, do not have a portrait defined.

This is just from a quick look, but by there could be a lot more errors/illogical choices, such as what relevance Marie Bonaparte has to Monaco. Please address these.

marijn211:
Such a range of characters only make sense to add if bound to focus tree content

PR #3162, opened 2025-05-01
May Bugfix Patch Day

BiccyThiccy:
[Plz put fixes in the patch notes as you make them](https://github.com/deliciousmods/1956_beta/wiki/May-2025-Patch-Notes)


BiccyThiccy:
@marijn2112 You can ignore the diff check for goe_shared_saadabad_branch.txt as it is just an imported vanilla file

marijn211:
> @marijn2112 You can ignore the diff check for goe_shared_saadabad_branch.txt as it is just an imported vanilla file

I assume you've made changes to it though? Otherwise there's no point in importing it

BiccyThiccy:
> > @marijn2112 You can ignore the diff check for goe_shared_saadabad_branch.txt as it is just an imported vanilla file
> 
> I assume you've made changes to it though? Otherwise there's no point in importing it

I'm not gonna lie I'm dumb as a rock, my bad. I think what I meant was that it was an imported vanilla file from 1.16.4. I think I changed it in a later commit anyways so my comment's irrelevant now. Just note that the file isn't intended to function atm, just there to get rid of errors.

BiccyThiccy (commented):
I did a quick diff-check on the German (and Iraq/Iran/Finland) changes like you asked Marijn, I'm not that great at eyeballing things but I didn't spot anything that looks wrong. The imported vanilla changes I incorporated elsewhere are much simpler thankfully as vanilla didn't shoehorn a huge list of changes like with Germany/India.

I also did hover checks on everything I did on this branch, and haven't found any problems.

common/characters/KOR.txt
```diff
@@ -499,6 +499,9 @@ characters={
 					has_government = fascism
 				}
 			}
+			visible = {
+				has_dlc = "La Resistance"
```
marijn211: @CertifiedEnjoyerOfSBG DLC checks like this should go in allowed blocks

common/national_focus/samoa.txt
```diff
@@ -707,8 +707,10 @@ focus_tree = {
 					has_autonomy_state = autonomy_wtt_imperial_associate
 					compare_autonomy_progress_ratio > 0.09
 				}
-				OVERLORD = {	#they are loosing, best time to leave this sinking boat
-					surrender_progress > 0.2
+				ROOT = {
+					OVERLORD = { #they are loosing, best time to leave this sinking boat
```
marijn211, line 711: This should be equivalent to what it was previously, if it threw an error, this shouldn't fix it

marijn211 (approved):
pull request meme

PR #3170, opened 2025-05-09
Finland Enhancements

Xaelorichero-02:
**Highlights**
- Merged R56 Finland tree with base game.


BiccyThiccy (approved):
Seems good. Error log is 2 lower than main and we cut out an error log spammer. 

Besides that, the gameplay bonuses should be pretty nice and it's always a good thing to cut out legacy content in favor of something newer.

PR #3182, opened 2025-05-31
Portugal - New Diplomacy Content, Many QOL/Balance Changes, and Improvements

BiccyThiccy:
**Additions:**
- New options for peaceful cooperation with Spain have been added to every path, which should hopefully help Spain and Portugal player-duos have more potential to have fun with one another, and open up interesting gameplay options.
- New focus has been added into the naval branch to develop Portugal's islands / treaty ports.
- One new advisor who led the Portuguese Legion until 1940.

Balance:
- Add penalties for full-annexing Spain via event (stability + political power)
- Rebalanced some of the advisors so that Portugal has less useless characters
- Focuses that require ideology support now continue if invalid (to avoid last-minute focus cancellations)
- Increased the penalty for full-annexing Brazil via decision (increased stability loss)
- Small buffs to the early-game section of the political trees, while nerfing late-game political power among a few other adjustments.
- Sprinkled some small military buffs throughout the focus tree, and gave more MIO bonuses for AAT owners.
- Lots of minor numbers changes to the various political decisions, which I think overall helps QOL and makes the decisions feel more different from one another.
- Changed the focus Modernize the Rural Regions -- no longer gives stability and a decision, but instead gives better buffs elsewhere.

Bugs / Modding Standards:
- Fixed the Democratic decision to core Guinea-Bissau not working
- Improved some localization that was not to current game standards (mostly legacy event stuff).
- Moved some stuff being named in event/focus-tree files into localization ones
- Moved a r56-added trait into the r56 traits file (was in vanilla).
Improved the wargoal AI a little bit (it was already acting fine but it should be more aggressive in appropriate situations).
- Every retire_character on Portugal checks for if they actually exist. Wouldn't cause issues usually but it would lead to errors upon the focus tree being reloaded, or through any other mean where the country can lose characters, so it's a necessary change.
- Changed event checks that remove state modifiers for if the modifier exists (would cause errors if it was gone otherwise)
- Checked some add non-aggression-pact event options for if a non-aggression-pact already exists (would cause errors if one existed otherwise)


marijn211 (commented):
Error log clean

localisation/english/PEM_l_english.yml
```diff
@@ -260,22 +260,22 @@ POR_elastic_defense:0 "Elastic Defense"
...
+ portugal.32.t:0 "Britain Accepts Our Request"
+ portugal.32.d:0 "Our request to join the Allies has officially been accepted by Britain. Our long-standing friends will fight side by side with us once again in the conflicts to come!"
+ portugal.32.o:0 "The Anglo-Portuguese Alliance stands strong!"
+ portugal.31.t:0 "Britain Reject Our Membership"
+ portugal.31.d:0 "Our request to join the Allies have officially been rejected by the British. With them being some of our longest-standing friends, we must wonder why they do not wish for our support."
+ portugal.31.o:0 "They should feel so lucky to have us."
+ news.704.t:0 "The Allies Grows"
+ news.704.d:0 "The influence of the Allies has grown today with Portugal officially joining the partnership. Only time will tell what this means for the future of global affairs."
+ news.704.o:0 "Portugal and Britain as allies? Tell me something new."
+ news.705.t:0 "The Allies Rejects [POR.GetAdjective] Ascension"
```
marijn211: I think this should be "Reject"? I guess if you mean the faction you could use this but it sounds unusual

marijn211 (approved):

PR #3190, opened 2025-06-03
Ridding the Chinese GUI, hopefully with no side effects

BiccyThiccy:
From what we know, GUI's are exceptionally bad for optimization due to the number of variables they constantly check for.

The China one has practically no gameplay use, although it is a neat system. Done with Spicy's backing, this PR removes it as well as a few references that must be removed to avoid errors.

I would propose getting this merged into the July update, although I doubt there are issues it would probably be unwise to shoehorn this in last minute.


marijn211 (changes requested):
"From what we know" is insufficient proof when you could do performance tests. I did one when this GUI was introduced that showed no difference

PR #3194, opened 2025-06-08
Speculoos south america

ghost:


BiccyThiccy:
Error log, really sick of telling you but you need to start checking your stuff before we're supposed to

ghost:
> Error log, really sick of telling you but you need to start checking your stuff before we're supposed to

I did check but I missed some﻿ :slightly_frowning_face:

ghost:
Checked the errors two times today, maybe there is one more error, but I cant find it with two times looking

ghost:
> Checked the errors two times today, maybe there is one more error, but I cant find it with two times looking

Checked the windows version and the lines are the same

ghost:
@BiccyThiccy Errors logs for all my 3 of my PRS should be clean now

common/characters/ARG.txt
BiccyThiccy, line 1: Please don't change vanilla focus tree checks. If you want to make an advisor require an ideology inside the RT56 tree, make a second instance of the character for those who don't have TOA.

common/characters/ARG.txt
```diff
@@ -307,6 +306,72 @@ characters = {
...
+			slot = political_advisor
+			idea_token = ARG_ernesto_alemann
+			allowed  = {
+				original_tag  = ARG
+			}
+			available  = {
+				has_government = democratic
+			}
+			traits  = {
+				liberal_journalist
```
BiccyThiccy: I would find a better role. A guy who gives democratic support is next to uselelss if you are already democratic. Makes sense?

common/characters/BOL.txt
```diff
@@ -53,17 +53,70 @@ characters = {
...
+			}
+			available  = {
+				has_government  = communism 
+			}
+			cost = 150
+			ai_will_do = { 
+				factor = 1		
+			}
+		}
+		#country_leader = {
```
BiccyThiccy: I would recommend keeping this role as fallback if Bolivia doesn't have many communist leaders. He can always be moved down the history file, but since Bolivia has the generic switch ruling party system, they can make multiple civil wars in a run which means you want as many leaders as possible to avoid a generic

common/characters/BRA.txt
```diff
@@ -973,7 +997,7 @@ characters = {
 		}
 	}
 	BRA_jorge_martins = {
-		name = BRA_jorge_martins
```
BiccyThiccy, line 976: Change his name = to 'BRA_jorge_martins_r56' and localize this new name please

history/countries/PAR - Paraguay.txt
```diff
@@ -44,19 +44,19 @@ if = {
 	recruit_character = PAR_amancio_pampliega
 	recruit_character = PAR_remigio_bastia_martin
 	recruit_character = PAR_francisco_esculies
-	recruit_character = PAR_Jose_Felix_Estigarribia
+	#recruit_character = PAR_Jose_Felix_Estigarribia #dupe
```
BiccyThiccy: If vanilla has a role for him, just rename ours to match vanilla

history/countries/PAR - Paraguay.txt
```diff
@@ -44,19 +44,19 @@ if = {
...
 	recruit_character = PAR_eduardo_schaerer
-	recruit_character = PAR_finance_graduate
+	#recruit_character = PAR_finance_graduate #already a finance advisor
 	recruit_character = PAR_fortification_armchair_general
 	recruit_character = PAR_aristides_rivas_ortellado
 	recruit_character = PAR_freire_gomez_esteves
 	recruit_character = PAR_ramon_paredes
 	recruit_character = PAR_mutshuito_villasboa_barboza
-	recruit_character = PAR_emilio_gardel
+	#recruit_character = PAR_emilio_gardel #already a finance advisor
```
BiccyThiccy, line 55: If he's real you should find another role instead of removing content

history/countries/PAR - Paraguay.txt
```diff
@@ -44,19 +44,19 @@ if = {
...
 	recruit_character = PAR_anselmo_jover_peralta
 	recruit_character = PAR_bernardino_caballero
 	recruit_character = PAR_pedro_duarte_ortellado
-	recruit_character = PAR_tomas_salomoni
+	#recruit_character = PAR_tomas_salomoni #already has a advisor with this trait
```
BiccyThiccy, line 59: If he's real you should find another role instead of removing content

BiccyThiccy (changes requested):
Please rename everything in localization files and get to my feedback, once it's done ping me for review

common/characters/ARG.txt
ghost: Manuel Galvez is a road char and the focus requirement that I removed is from RT56 tree

history/countries/PAR - Paraguay.txt
ghost: Both vanilla and road version are the same

history/countries/PAR - Paraguay.txt
ghost: There no info on them besides there ministerial post, then there are other advisors that he those traits

history/countries/PAR - Paraguay.txt
ghost: There no info on them besides there ministerial post, then there are other advisors that he those traits

BiccyThiccy (approved):
Seems fine and I see no new errors

PR #3198, opened 2025-06-22
Speculoos companies

ghost:


common/ideas/britain.txt
```diff
@@ -2719,7 +2719,26 @@ ideas = {
 				production_speed_steel_mill_factor = 0.10 
 			}
 		}
-		ENG_vulcan_foundry = { # made austerity trains
+		#ENG_vulcan_foundry = { # made austerity trains #alreadt a train mio
```
marijn211: So?

marijn211 (changes requested):
Error log is clear, but I find it very hard to judge the PR, I think there should either be comments or a PR description to make clear why things are swapped or removed

E.g. does a company really have to be swapped because it's foreign?

It's a bit like mapchanges without the focus tree being changed, are we really gaining anything by changing these things?

marijn211 (approved):
Comments look good

PR #3202, opened 2025-07-02
Bugfix July25 Patch Day

BiccyThiccy:
Use these as you go please:

Patch Notes: https://github.com/deliciousmods/1956_beta/wiki/July-2025-Patch-Notes

Please briefly hover over any focus you modify, and if you make significant additions check the entire tree and if you find any, fix them or report them; if there are none, add the country here: https://docs.google.com/document/d/1w2dStPoE_LBmcskbwXRJ5pKvNAj06W3awRq6e_ssPP4/edit?tab=t.0


common/on_actions/r56_on_actions.txt
BiccyThiccy, line 1: Shouldn't this go in the focus_tree_selection file?

Xaelorichero-02: No. This is a compability check when you set Asia free.

common/characters/AFG.txt
```diff
@@ -293,7 +283,10 @@ characters = {
 			cost = 100
 			available = {
 				NOT = {
-					has_completed_focus = AFG_form_weekh_zalmian
+					if = {
+						limit = { has_global_flag = afg_dlc_focus_tree }
```
marijn211, line 287: These halve empty if-else statements look quite odd

common/characters/AFG.txt
```diff
@@ -522,7 +505,10 @@ characters = {
...
```
marijn2112, line 509: This looks odd

common/characters/AFG.txt
```diff
@@ -765,22 +731,26 @@ characters = {
...
 						has_country_leader = {
 							character = AFG_amanullah_khan
 							ruling_only = yes
 						}
-						has_country_flag = AFG_approached_haji_mirzali_khan
-						NOT = {
-							has_country_flag = AFG_dissatisfied_haji_mirzali_khan
+					}
+					else = {
+						AND = {
```
marijn211, line 745: Doesn't have be adjust but be aware that inside `if` statements, `AND` is the default again, even if the scope outside it was `OR`

common/characters/AFG.txt
Xaelorichero-02: This looks completely normal

BiccyThiccy: > These halve empty if-else statements look quite odd

I don't think we have an if check for the 56 tree, it looks a bit weird but having a NOT={ check for the dlc tree wouldn't look quite right either imo

It would look more normal if it was filled with stuff in the  if block

marijn211: > This looks completely normal

The highlighted part does, but there being nothing in the if-statement just below and only the else looks weird

marijn211 (approved):

PR #3205, opened 2025-07-04
Fixing vanilla bug(?) with Dardanelles Strait Rights Being Blocked While Bosphorus is Open

BiccyThiccy:
I checked vanilla for references to the deleted flag, and this is the only one besides localization that I could find.

This commit just replaces it with the Bosphorus one, if this is an inappropriate fix let me know


marijn211 (commented):
What if this is intentional to avoid ship trapping?

BiccyThiccy:
My theory is that it's an oversight though because the flag they did opt for was completely unused.

Ships can technically be trapped anyways if control changes mid war, so if it's a rare but possible scenario, I think guaranteeing your ship can get through the Dardenelles if you have country_flag passage to the Bosphorus is preferable to a rare thing happening less.

This can just be delayed to September though because it is kinda last second, maybe I can look more and find another fix

marijn211:
> > What if this is intentional to avoid ship trapping?
> 
> I think it is a lot better to have trapped ships than being unable to go through the Bosporus

The flag is only for disabling passage, not enabling

To be clear, the passage definitely has cut content and bugs in vanilla, but this could be intentional

BiccyThiccy:
@marijn2112 Should we look into this further or should I just close this branch?

marijn211:
Personally Id say close it since they should probably use the same behavior

PR #3216, opened 2025-07-25
Hellas

V4ggos:
Added new stuff For Greece


common/characters/GRE.txt
```diff
@@ -114,102 +101,121 @@ characters={
...
-			}
-		}
-		advisor={
-			slot = political_advisor
-			idea_token = GRE_ioannis_metaxas_bftb
+		instance = {
 			allowed = {
-				original_tag = GRE
 				has_dlc = "Battle for the Bosporus"
+				has_game_rule = {
```
BiccyThiccy, line 107: Game rules don't work on allowed blocks in character files.

The way this was done was appropriate but if you really want to, let's say for the purpose of the character being transferrable - create the character role in the on_action for the dlc selection instead of this.

common/national_focus/r56_greece.txt
```diff
@@ -2499,21 +2472,21 @@ focus = {
 	focus = {
 		id = GRE_reinforce_4th_of_august_regime
 		icon = GFX_focus_GRE_4th_august
-    	text = GRE_reinforce_4th_of_august_regime
+		text = GRE_reinforce_4th_of_august_regime
```
BiccyThiccy, line 2475: Fyi these text = can probably be removed, if you're gonna fix indenting just cuz, you can pull stuff like that also

common/on_actions/r56_focus_tree_selection_on_actions.txt
```diff
@@ -1104,7 +1103,7 @@ on_actions = {
 						election_frequency = 7
 						elections_allowed = yes # none after offset 1
 					}
-					#promote_character = GRE_ioannis_metaxas - has double roles and he shall definitely not be neutrality leader from game start - Fantom
```
BiccyThiccy, line 1107: Why remove this from the comment? It's there in case someone has a question so they aren't guessing who they have to ask.

PR #3225, opened 2025-08-03
Biccy Tweaks Patch Day

BiccyThiccy:
**Things Not in the Patch Notes:**
- POR: Small numbers changes have been implemented, making the process a bit faster to flip fascist in the focus tree.
- GRE decision file -- non-vanilla content has been moved into a new GRE_r56 file, stuff that can go in allowed blocks has been taken out of visible/available ones, a few bugs have been fixed and file has been reorganized to be more similar to vanilla.

**Patch Notes:**

**Adjustments:**
- On Portugal's Road to 56 focus tree, the Communist/Democratic coup may now be performed peacefully if there is less than 30% authoritarian support in the country.
- Incorporated many vanilla changes made to Italy since the launch of By Blood Alone.
- [BftB] Incorporated many vanilla changes made to Bulgaria since the DLC's launch.

**Bugfixes:**
- Fixed decision "Take Control of the Japanese Mining Companies" never appearing on Vietnam.


marijn211 (commented):
Diff seems fine, though I think we should have a discussion about #r56 comments and restoring empty code

BiccyThiccy:
I can be less aggressive with comparison software going forward if it's a problem, I prefer to be as specific as possible to avoid confusion when doing maintenance on these files - it hasn't been getting done, not that this was probably why but it should make porting vanilla changes a faster process going forward here.

I can try to undo some of it if you want but I don't think there's much risk as most focus tree issues are easy to spot compared to more-hidden stuff like events or back end code.

marijn211 (approved):

marijn211 (approved):

PR #3220, opened 2025-07-28
Xaelor Septemberprogramm

Xaelorichero-02:
**Highlights**
- null

**Additions**
- Walloon and Flemish trees are now based of the RT56 generic tree.
- Added Sweden-Hungary and the Inuit Union as formables.

**Adjustments**
- Removed the duplicate Sikkim tag – `SKM`, Guyana tag – `GBR`, Don Cossack tag – `DOC` & Greenland tag – `GRL`.

**Bugfixes**
- null

**GFX**
- Readded many TfV and DoD portraits.
- Improved a few Polish portraits.

**Localization**
- null

**Credits**
- Please check out our 'Credits.txt' links and order of battles for more information about the people, mods and resources that helped develop this project, as well as for opportunities to learn about the related history.

PR #3241, opened 2025-09-01
Quebec tag

Xaelorichero-02:
**Additions**
- Ernest James Renaud will transfer to Quebec after its independence.

**Adjustments**
- Removed the duplicate Quebec tag – `QBC`.


Xaelorichero-02:
Originally part of #3220

marijn2112 (changes requested):
Those country name things (r56_countries_l_english) seem better than vanilla, please more them to a file in the replace folder instead of deleting them

marijn2112:
Unrelated as the mistake predates this PR, but I'm fairly certain some of the events are using `QUE.GetNameDef` where they should be using `QUE.GetNameDefCap` instead.

Xaelorichero-02:
> Those country name things (r56_countries_l_english) seem better than vanilla, please more them to a file in the replace folder instead of deleting them

Ik u streaming now but "more" makes zero sense in this sentence.

marijn2112 (approved):
Seems to work fine, I'll merge this after seeing if I can include a few localization bugfixes

PR #3246, opened 2025-09-08
E bugfix october

### LINKED ISSUE #3240: Polish focus tree branch "Prepare for the Inevitable" seems to have bug with its prerequisites.
The right most focus tree of Poland is meant to be used when the country is likely to be capitulated, or already capitulated. The first focus of this branch has prerequisite of having "more" than 50% of surrender progress, and this seems fine. Other focuses however, has "less" than 50% of surrender progress. This difference causes player unable to go down this branch any further during war, at least until player surrenders. 

Image

Image

According to old paradox wiki's document, this bug could also be found in early days of vanilla's reworked polish focus tree.

https://forum.paradoxplaza.com/forum/threads/hoi-4-poland-prepare-for-the-inevitable-in-new-focus-tree-is-broken.1500264/

It would be awesome if this bug gets fixed in some way so we can enjoy some timeline where poland standing against germany in harsh conditions.

e-ee-ee:
Various bugfixes for China and Warlords (including Mengukuo) and a capital picker.

## Changelog:
**Highlights**
- null

**Additions**
- Added an option to choose the new Chinese capital for Warlords.

**Adjustments**
- Army/Air/Navy Old Guard Chiefs now grant extra political power.

**Bugfixes**
- null

**GFX**
- null

**Localization**
- Custom scripts can be rendered in the election GUI.


common/decisions/BEL_HOL_decisions.txt
```diff
@@ -273,7 +274,8 @@ form_nation_mod_category = {
 		icon = generic_army_support
 
 		allowed = {
-			original_tag = HOL		
+			original_tag = HOL
+			original_tag = LUX
```
marijn211: You forgot an OR

BiccyThiccy: Ah crap 😭 I'll fix it shortly, good catch

common/national_focus/china_nationalist.txt
```diff
@@ -17392,6 +17392,23 @@ focus_tree = {
...
+			}
+			every_other_country = {
+				limit = {
+					is_subject_of = ROOT
+				}
+				set_autonomy = {
+					target = AST
+					autonomous_state = autonomy_free
+					end_wars = no
+					end_civil_wars = no
```
marijn211: This is the unedited example from the wiki

marijn211: @e-ee-ee

common/national_focus/china_nationalist.txt
```diff
@@ -12178,14 +12177,14 @@ focus_tree = {
 					limit = { 
 						free_building_slots = {
 							building = arms_factory
-							size > 1
+							size > 2
```
marijn211: This is unusual, usually you check size 1 smaller than what you're adding, as >2 means at least 3 free slots, but you're adding only 2

e-ee-ee: Alrighty

marijn211 (approved):
Ill merge this in around 12-14 hours, probably after doing another historical+performance check run

PR #3247, opened 2025-09-13
Defending Earthrealm

katsuragiyy:
**Highlights**
- null

**Additions**
- null

**Adjustments**
- null

**Bugfixes**
- null

**GFX**
- Improved multiple Venezuelan portraits.

**Localization**
- null

# UNRELATED / NOT IN PATCH NOTES
- Converted most Filipino content to have r56 tags for a smoother integration of the DLC.
- Merged the Argentine idea file.


katsuragiyy:
I did not expect us to move this fast towards DLC integration
since this contains some things for the PHI, and most of my attention is going towards my own fork, i might as well let this be merged as soon as possible (even tho i had hoped to do more)

PR #3249, opened 2025-09-15
A GUI bug related to the 'China General Election' is causing question marks (?) to be displayed for users on the Simplified Chinese version.

zhongtiandang:
Hello, this is a bug report from the translation team regarding a font rendering issue in the Chinese locale. We've used the Traditional Chinese version for this report as it supports two different fonts for testing. As shown in the attached images, text renders correctly using the specified font in English mode (Figure 1), but displays incorrectly when using the native Paradox font in Chinese mode (Figure 2). I think my analysis suggests the Chinese locale may not support the hoi_16 and hoi_20 font definitions. After consulting, we resolved the issue by modifying these to hoi_16mbs and hoi_20b respectively, which resulted in a correct display as shown in Figure 3. We request that the development team implement this fix if it does not adversely affect the English version. Additionally, we have attached a separate file with corrections for other localization errors.


marijn211:
Could you try to fix the diff showing the entire files as being changed?

zhongtiandang:
> > 您能否尝试修复显示整个文件已更改的差异？
> 
> 检查行尾是否为 CRLF 且编码器是否为 UTF-8（本地化的 UTF-8 BOM） 作为参考，您可以在 Visual Studio Code 中更改它。 图像

My fault. Someone told me that changing the encoding could fix the question mark bug but helpless. Do you need me to re-upload it?

marijn211:
If that fixes it, feel free to leave it like that, otherwise its more convenient to check with the original line endings, as the diff tab of the PR will be more usable

zhongtiandang:
wait a second……I am sure my files are edit with UTF-8 and CRLF, But  the original files are same too？
5afc66e7-45a0-438c-b507-a83c3a48fdd7
Or you can try to raise another PR on my behalf, anyway, you already understand what I mean

Xaelorichero-02:
I have discovered the issue. E used non existing fonts. So I remade your edits with the up to date file on our development branch.

zhongtiandang:
Okay, appreciate it, then can I close with the comment?

PR #3256, opened 2025-10-05
Reintroduce jet frame

marijn211:
Splits Jet engine III into a separate special project while restoring the jet frame.
Improves special project images for jet engine I and jet engine II while recycling the old jet engine II one for jet engine III.
Jet Engine II icon now uses the vanilla one which looks nicer.


BiccyThiccy (approved):
No errors or any broken code, seems fine in-game too

PR #3257, opened 2025-10-05
Waffle september bugfix1 (should probably be october) - canada tweaks

Thinking-waffle:


common/characters/CAN.txt
```diff
@@ -933,7 +933,7 @@ characters = {
 			traits = {
 				air_pilot_training_3
 			}
-			cost = 150
+			cost = 200
```
marijn211: Any particular reason for the inconsistent cost @Thinking-waffle ?

Everything else looks good

Thinking-waffle: It's a mistake but a mistake born out of a frustration of the efficiency of certain traits compared to their usefulness. It's both the best trait for the guy in his later years and still quite costly to spend on him alone.
At the same time I should have left a note.

marijn211 (approved):

PR #3259, opened 2025-10-08
Silesian Voicelines (and other contributions)

InfraTheIdiotic:
**Additions**
- Unique voicelines for Silesia recorded by my friend RazZorInc, transcriptions and meanings are in the metadata of the sound files.
- Generic advisors for Silesia through `history/general/generic_advisors.txt`.
- A Western Poland Fragmentation Status game rule to release Silesia and Kashubia. A new category was made so the east can be adjusted seperately. Releasing the west as independent will give Danzig to Germany, as releasing the east already does.

**GFX**
- Changed the Silesian fascist flag to the one the Union of Upper Silesians (fascist party) actually proposed.
- Added some alternate ideology flags for Mauritius, Dominica and Saint Lucia.

**Localization**
- Adjusted the localisation of (Eastern) Poland Fragmentation Status to distinguish it from the west.


marijn211 (changes requested):
Seems more sensible to just merge the two fragmentation rules and have 2 extra options that decentralize/fragment both the East and the West

Instead of telling people to not enable Remove Some Releasables, just adapt the script of that gamerule

The releasables are also very weak, if KSH and Silesia are to be released by gamerule it seems more sensible to give them Danzig and German Silesia

Error log currently clear

marijn211 (commented):
Voicelines and flags seem fine

InfraTheIdiotic:
Remove Some Releasables not being overrode was one I was more uncertain about, I only didn't do it because others don't and the loc for it acknowledges that but I can sort it out for this, I'll begin sorting all this out now, thanks for the feedback.

marijn211 (approved):

PR #3262, opened 2025-10-11
Alter Tunganistan balance and border

marijn211:
This solely reassigns provinces and shouldn't break savegames


marijn211:
Already confirmed there's no error log so should be ready, unedited population shouldn't be an issue, no cities included in the tiles I moved

TheBrit7 (approved):
Looks good boss.

PR #3270, opened 2025-10-18
Three bugfixes

marijn211:
Undo the random argument which actually makes the lootboxes less random, this sadly means itll be repetitive again in multiplayer


BiccyThiccy (approved):
Can confirm it fixes the focus tree on_action error

PR #3271, opened 2025-10-27
Red November's 99th Anniversary

KrasnaMerah:
PKI Path
* Tan Malaka is added to the mod
* "PKI" focus would make Amir a communist-boosting advisor
* "Reorganize the PKI" would make Musso the leader of Underground PKI (he was here since December 1935) instead of going back to the Soviet Union
* Semaun's initial trait is changed to "Former Railroad Union Leader" and would return from the Soviet Union after the revolution is finished
* Jokosujono would be unlocked by "Reorganize the PKI" instead of "PKI" focus (he was in hiding in 1936)
* Added Party Congress events (3rd to 7th)

MISC Indonesia
* Hok Hoei Kan added back to the mod

MISC
* Philippines: Tan Malaka's trait is boosted and also has a non-LaR version (NCNS only)

Intended:
* More advisors
* Islamist and Fascist independence event chain
* More MIOs


KrasnaMerah:
Early review for now. Hopefully i can add more stuffs.

marijn211 (changes requested):
I dont expect major conflicts but just to be sure lets hold this until after DLC patching

common/characters/INS.txt
```diff
@@ -1184,6 +1205,55 @@ characters = {
...
+			}
+			expire = "1965.1.1.1"
+			id = -1
+		}
+		advisor = {
+			slot = political_advisor
+			idea_token = INS_tan_malaka
+			allowed = {
+				original_tag = INS
+				has_dlc = "Man the Guns"
```
marijn211: Doesn't seem necessary? His trait doesn't affect anything MtG related

Tan Malaka has also been added on the Philippines in SEA, you could instance this character if you like the DLC portrait more

common/country_leader/r56_traits.txt
```diff
@@ -3344,6 +3344,15 @@ leader_traits = {
 			value = -200
 		}
 	}
+	dislikes_stalin = {
+		random = no
+
+		ai_strategy = {
+			type = antagonize
+			id = "SOS"
```
marijn211: SOS I believe only targets Stalinist USSR if they're in a civil war

common/characters/INS.txt
KrasnaMerah: This is a leftover from EIR, as i'm locking several characters behind MtG.

Gotta delete this since MtG has been integrated

common/country_leader/r56_traits.txt
KrasnaMerah: Okay, i'm on it

common/country_leader/00_traits.txt
```diff
@@ -17242,6 +17242,17 @@ leader_traits = {
...
+		boost_ideology_mission_factor = 0.15
+		compliance_gain = 0.02
+	}
+
+	PHI_indonesian_communist_exile_no_lar = {
+		random = no
+		sprite = 13 
+
+		communism_drift = 0.05
+		boost_ideology_mission_factor = 0.15
```
marijn211, line 17254: I assume this is a LaR-exclusive modifier, though please confirm

Xaelorichero-02: It isn't. If you don't have it, it works on the old diplo actions to increase party popularity.

marijn211 (approved):

PR #3283, opened 2025-11-21
NCNS Map Fixes

katsuragiyy:
>Corrected the terrain for China.
>Fixed weird vp positions and similar instances.


marijn211:
Can you explain why the unitstack diff is so big?

katsuragiyy:
@marijn211 because of the absurd amount of provinces that had to be randomized (~150 provinces or somewhere around that range i'd say)

marijn211 (approved):

PR #3287, opened 2025-11-22
Fixing a few things - Waffle december 1

Thinking-waffle:
added coal to coal focus for Tuva.

added the coal consumption reduction from dams.

changed a few dams located in the mountains into mountain dams.

Added a tiny amount of coal and/or coal mining decisions in relevant regions of Africa (Katanga, Tanzania, Rhodesia/Zimbabwe) 

Adding a few mining unique focuses for independent Katanga, for funzies, flavor... (and as a nod to my grandpa)


common/national_focus/katanga.txt
```diff
@@ -0,0 +1,1690 @@
+#####-----> Rt56 Tuvan Tree <><> <><> <><>
```
marijn211: Rename

history/states/889-Elisabethville.txt
```diff
@@ -1,50 +1,45 @@
...
+		buildings = {
+			infrastructure = 1
+		}
+
+		add_core_of = COG
+		add_core_of = KAT
+
+		victory_points = {
+			191 2 # Elizabehville
+		}
```
marijn211, line 36: It seems you removed one of the victory point? Is that intentional? In general this diff is broken

Thinking-waffle: that's weird, assume a no. I am not sure what's wrong. Elizabethville is still there. It seems fine. Maybe I a missing something but I launched it and it ran okay. Except that Elizabethville is outside of the border (again?)

marijn211: The other victory point I mean, 1516

marijn211 (approved):

PR #3297, opened 2025-12-02
fixing stuff if I can (check description marijn)

Thinking-waffle:
- Fixing state problems in malayan resources faction goal.
- Fixing state problems relative with the Atlantikwall and the German coastal AA defenses. (focus, decisions, tooltips)
- Moved the Saguenay aluminium to the state with the Saguenay city.

*GFX*
- Used new "Indian militia" (without turban) model for the regular infantry of a bunch of indian minors, added the Turban one to others. I didn't change the model for cavalry, motorized etc. as the risk of errors would be too great in a beta for a lot of tags.
#####

Notes: 

- [x] - Sabah "North Borneo" 916 in rt56 has 34 rubber in the base game. Problem Github doesn't show me all the state files. Not sure if that rubber has been removed elsewhere.
- [ ] - state name changes for the Malaysian part of Borneo should be considered.
- [x] - Singapore is not part of the objectives due to the lack of resources in the version I am suggesting, but it could be.
- [x] - Some states in Quebec will have  to change name but that will probably wait for later.


marijn211:
image
Heres the resources if that answers your first question


Singapore not being part of the resource-related goal seems good, since it indeed has no resources

marijn211 (approved):

PR #3298, opened 2025-12-02
December bugfixing 2

marijn211:


common/national_focus/poland.txt
```diff
@@ -415,17 +415,35 @@ focus_tree = {
...
+						free_building_slots = {
+							building = steel_mill
+							size < 2
+						}
+					}
+				}
+				add_resource = {
+					type = steel
+					amount = 8 #r56 was 24
+					state = 762
```
marijn211, line 438: This should be an if-else @Xaelorichero-02 , this order of operation means there might be sufficient room for the buildings, causing them to be placed, which might in turn reduce the building slot amount enough to also trigger the 2nd reward.

interface/unitleaderwindow.gui
marijn211, line 1: I checked and both of these file overrides were in fact not useless, Xaelor, please don't restore them though as the fixes they were doing were flawed and I'm looking into improving them myself

Thinking-waffle (approved):
good luck.

PR #3303, opened 2025-12-06
Restore building speed modifier overlay UI

marijn211:


marijn211:
Already confirmed this to be error-log-free the only issue I can imagine is that it lags multiplayer if someone spams the button, in singleplayer it works quite smoothly but it does do a little math to order the menu

Thinking-waffle (approved):
keep it up.

PR #3305, opened 2025-12-08
Bugfixing for January 2026 Patch

ilkober:
All the necessary (and not) bugfixes and tweaks


marijn211:
@ilkober Could you greatly limit border wars for Mongolia? Especially for AI Mongolia. Its really annoying when you're playing Japan

BiccyThiccy:
@e-ee-ee I think one of your commits has added an error regarding the new shared branch

localisation/english/TIB_l_english.yml
```diff
@@ -439,8 +439,8 @@
  himalaya_tibet.43.a:0 "How disappointing."
 
  himalaya_tibet.44.t:0 "Stalemate at [434.GetName]"
- himalaya_tibet.46.desc:0 "Our military forces have not been able to advance in the battle of [434.GetName]. We have fended off many enemy attacks but our offensive charges and morale were not good enough to push through the enemy lines. However, there is still much that we can learn and information we can gather from these skirmishes..."
- himalaya_tibet.46.a:0 "A waste of blood and tears..."
```
marijn211, line 443: This still seems broken, this probably doesn't have to be undone, but in the events files, there seem to be several references to non-existent loc that need to be worked away

common/national_focus/sudan.txt
marijn211, line 1: Can you elaborate @Xaelorichero-02 ?

marijn211:
Other than those 2 comments the diff seems fine

common/national_focus/sudan.txt
marijn211: Ah hidden in that sense, good to know

marijn211 (approved):
lll try include a few lastminute fixes and then merge to make place for a new bugfix PR

PR #3306, opened 2025-12-09
waffle doing some stuffs

Thinking-waffle:
- Cambodian characters (death of king Sisovan may need an event)
- King of the Yete for Katanga, tweaked their name list
- change in the generic portraits ( using some MtG portraits etc.)
- remove duplicate MIO for Korea
- fixed trait of a Pakistani general (first step of a review of Indian characters)
- made use of alternative fall of Athens pictures
- added "african" fall of istanbul picture
- added characters for the Chinese warlords.

WARNING: despite adding characters there are still holes in their list of advisors, weakening them significantly, the addition of generics may be necessary for naval/air advisors in the short term.
That content also has to be checked with and without the dlc, even if I am fairly confident.
Because of the relationship between the ma and the confusion i had to sort them I used r56_on_action to distribute some of them among the new warlords while they are all written in XSM. It's not an ideal system but it saved me a headache in the very short term.
ideally we would have to identify the army names attributed to the warlords in order to search for some warlord officers in the database.


marijn211:
Error log is clean with my commit, will check diff etc. later

marijn211 (changes requested):
Error log now includes a few major issues including:
A bracket error in SICs file around line 187
Usage of maoism ideology which we dont have
[19:39:27][no_game_date][database_scoped_variables.cpp:266]: invalid database object for effect/trigger: JAP_silver_wheels. use var:var_name to explicitly use variables in effects/triggers
[19:39:27][no_game_date][database_scoped_variables.cpp:266]: invalid database object for effect/trigger: JAP_lighter_tanks. use var:var_name to explicitly use variables in effects/triggers
[19:39:37][no_game_date][character_template_database.cpp:242]: Multiple character have the tag XSM_ma_chengxiang
[19:39:37][no_game_date][character_manager.cpp:723]: Several characters have the name XSM_ma_chengxiang

common/characters/GDC.txt
```diff
@@ -0,0 +1,198 @@
...
+				attack_skill = 2
+				defense_skill = 2
+				planning_skill = 2
+				logistics_skill = 1
+				legacy_id = -1
+			}
+		}
+	}
+	GDC_ma_binghong = { # chief of logistics in guandong 36-38
+		name = "Ma Binghong"
```
BiccyThiccy, line 52: I would recommend getting into the habit of making your name= stuff into proper locstrings that are named in a localization file. Translation submods that use different alphabets can't do anything to fix this

history/countries/GXC - Guangxi.txt
```diff
@@ -83,6 +83,19 @@ recruit_character = GXC_zhu_jiaxun
...
+recruit_character = GXC_wei_yunsong
+recruit_character = GXC_xu_qiming
+
+if = {
+	limit = {
+		NOT = {
+			has_dlc = "No Compromise, No Surrender"
+		}
+	}
+	recruit_character = GND_yu_hanmou
```
BiccyThiccy, line 96: I know I just did the changes so you didn't code this wrong, but just remember that Guandong, Hebei-Chahar, Shandong and Northeastern Army is going to exist for everybody - so make sure you're not recruiting characters for them onto GXC/CHI when the DLC is not loaded

marijn211 (approved):
Net negative errors, though would like a code comment why YUN_liu_zhao got commented out

PR #3307, opened 2025-12-17
Australia update

Tjscott97:
Focus tree expansion across economic, army, navy and airforce branches adding 14 new focuses.

Approx. 20 economic ideas merged into one dynamic modifier.

New evolving army idea.

New navy idea.

Minor state resource changes.

Commonwealth Aircraft Corp MIO modified to be able to produce CAS as well as a new unique trait.

Several less valuable focuses across political branch have been shortned and 'Control the ABC' focus has been moved to the shared commie-fascist branch giving a 5% boost to either fascism or communism depending on what branch has been taken.

Multiple focus icons have been updated to more fitting newer icons that have been added to the game since focus tree was made.


history/states/520-Northern Territory.txt
```diff
@@ -3,7 +3,7 @@ state={
 	id=520
 	name="STATE_520"
 	resources={
-		steel = 22
+		#steel = 22 #there was no steel production in northern territory at this time
```
marijn211: Could maybe add to the comment that you moved the production to 517

history/states/517-Victoria.txt
```diff
@@ -3,7 +3,7 @@ state={
 	id=517
 	name="STATE_517"
 	resources={
-		steel = 8
+		#steel = 8 #there was no steel production in victoria at this time
```
marijn211: Idem

common/military_industrial_organization/organizations/AST_organization.txt
```diff
@@ -132,6 +132,23 @@ AST_com_aircraft_corporation_organization = { #
 	}	
 }
 
+AST_R56_com_aircraft_corporation_cas_organization = { #Historically built CAS aircraft
```
marijn211: Adding a company twice because they built multiple types isn't consistent, try to find a type/modify the upgrade tree so the MIO can apply to both types they produced

marijn211:
A few things are broken, to detect surface-level stuff, launch the game with -debug launch argument and you'll get a debug log with errors

marijn211 (changes requested):
See above

common/on_actions/r56_on_actions.txt
```diff
@@ -21367,6 +21370,28 @@ on_actions = {
 				}
 				DEN = { set_country_flag = den_lost_core_state }
 			}
+			if = {
+				limit = {
+					FROM = { tag = AST } 
+					NOT = { ROOT = { tag = AST } }
```
marijn211: If you want to exclude civil wars, use original_tag instead of tag, FROM and ROOT can't both be the tag AST

marijn211:
Diff and error log look good aside from that thing above, would only need a glance in-game and then should be good

marijn211:
Please indent properly in future code

marijn211 (approved):

PR #3311, opened 2025-12-24
More bugfixing for January

marijn211:


common/characters/JAP.txt
```diff
@@ -3559,9 +3559,13 @@ characters={
...
-			army={
-				large = GFX_kr_portrait_asia_generic_land_4
-				small = GFX_kr_portrait_asia_generic_land_4_small
+			civilian = {
+				large = GFX_portrait_JAP_jinzaburo_masaki
+				small = GFX_portrait_JAP_jinzaburo_masaki_civilian_small
+			}
+			army = {
+				large = GFX_portrait_JAP_jinzaburo_masaki
+				small = GFX_portrait_JAP_jinzaburo_masaki_military_small
```
marijn211: Thats a DLC-only portrait on a character also available to non-DLC players @CertifiedEnjoyerOfSBG 
(I think Biccy edited him to be allowed with has_completed_focus = JAP_support_the_kodoha_faction which is a non-DLC focus)

Would need to be instanced, potentially also import his army chief role which we didn't port

katsuragiyy: aa i assumed he was dlc locked. nvm then

katsuragiyy (approved):

PR #3329, opened 2026-01-20
Zoloto mega branch

Gold3nBat:
I'll add the Monaco stuff for the Monaco Branch later, needs to be updated, it's better for me to send in the popes & GUK characters and not rush into Monaco stuff


history/countries/PAP - Papal States.txt
```diff
@@ -134,31 +134,45 @@ set_politics = {
...
+recruit_character = PAP_eugen_tisserant
+recruit_character = PAP_luigi_maglione #vatican secrtariat of state 1939-1944
+###===========================> Any
+recruit_character = PAP_francis_spellamn
+recruit_character = PAP_francesco_marmaggi
+recruit_character = PAP_antonio_riberi
+###==================================================> Chiefs
+recruit_character = PAP_george_von_sury_d_aspermont
+###==================================================> High Command
+recruit_character = PAP_umberto_aebischa #Swiss Captain
```
marijn211, line 172: In most cases you'd want to switch the nationality of characters instead of duplicating them to avoid situations where the characters fight themselves

marijn211 (changes requested):
Error log

marijn211:
Will close because development is paused

Note to self: Papacy currently cant spawn without console I believe, make that a new option for the Italy fragmenting gamerule

This also means currently the concern of generals fighting themselves won't occur without console but the cause of that is more important to fix than preventing it happening

PR #3331, opened 2026-01-26
Dominions now notified of UK investment

katsuragiyy:
Additions -- >
-- Once the UK invests into one of the dominions through their Imperial Federation branch, they get notified of the investment via event.

this has been rotting for the past week and i have nothing else to add so might as well get this merged


marijn211 (approved):
Fixed erroneous fire_only_once, seems good otherwise

PR #3336, opened 2026-02-04
Andorra gfx

TheBrit7:


marijn211 (changes requested):
Still remnants of a conflict in the two ideas-related .gfx files

BiccyThiccy:
I feel like we should be incorporating those icons *somewhere*, we shouldn't make the mod larger for no reason

marijn211:
Can be merged into branches that need this GFX

PR #3342, opened 2026-02-22
New shared bugfixing march 2026

BiccyThiccy:
Please put fixes and changes [here](https://github.com/deliciousmods/1956_beta/wiki/Patch-Notes-March-2026)

Our number one priority is integrating changes from vanilla's most recent patch. Please go through [this list](https://docs.google.com/document/d/18ONemHczcNjxlLstH1bMzODuUO5j1bPjUh0DV-L3Y8I/edit?usp=sharing) and incorporate whatever you can.

CHECK ERROR LOG BEFORE VS WITH YOUR CHANGES


marijn211:
@BiccyThiccy Feel free to move the commit to a different PR but Im gonna revert the allowed block changes for now

In theory the original_tag/tag simplifications you made in allowed are not functionally identical
Similarly, you moved VIC into allowed blocks, but VIC is a tag alias, so it could in theory (not sure if its the case) be that the allowed is checked on country spawn, and that a few microseconds later the flags are set that turn that country into VIC
Same is I suspect could be an issue with the SPC tag you moved into allowed

I also obviously do performance tests and there are some already-optimized/unoptimizable decisions that tend to dwarf the impact of the allowed fixes

marijn211 (approved):

PR #3344, opened 2026-03-02
Dutch Reworks Split

DutchPlanDerLindeSus:
- Hundreds of updated and new portraits
- Reworked historical railways and supply hubs (limited tags)
- Fixed UK southern shore gfx issue by updating colormap_rgb_cityemissivemask_a
- Historical population rebalancing for multiple tags
- Historical railway, infrastructure, VP and building slot rework for multipel regions/countries


marijn211 (changes requested):
Error log, state files also have edits you didnt describe in your patch notes

marijn211:
If the errors are because the PR is split in two, please split it differently, see previous comment too, I will do a quick glance over some portraits

marijn211:
If I see correctly the portraits are added but not implemented in the PR now, that's also an issue

PR #3351, opened 2026-03-12
Partial Czechoslovakia rework

TheBrit7:
There still may be some lingering errors, but most should be fixed. Please let me know.


marijn211:
Please check conflict

marijn211:
Error log is indeed clean and branch now up-to-date, I thinking about whether we should revert the CZKE->CZE changes just to make adding Czechoslovakia pack a lot more easy, even if it's super inconsistent and I'd have otherwise changed it too at some point

marijn211 (changes requested):
Form Fast Divisions has a broken tooltip

There either isnt- or AI isnt doing- content to guide fortline expansion, leaving AI with 0/1 fort level when Demand Sudetenland happens, instead of 3~ average previously

Everything else appears fine at first glance, though before accepting should take another look if any removed focuses would fit nicely into the reworked eco/military focuses, I noticed the post-war ones for example are gone, might be easy to reinsert

marijn211 (approved):

PR #3355, opened 2026-03-21
Make province renaming only apply to ones with victory points

marijn211:
Inspired on how KR does it


BiccyThiccy (approved):
Haven't done a readthrough of code since I don't fully understand it, but based on a large sampling of stuff it seems like this fixes the bugs with the tool and works correctly in general

PR #3360, opened 2026-03-29
RAJ - Army/Navy/Air Rework + Other Adjustments

BiccyThiccy:
**Interesting Country's:**
- Swapped Khotan Clique for British RAJ and rewrote the description to be more vanilla-quality.

**India Politics:**
- Added line-skips for quality of life and changed the order which some effects are displayed.
- Tweaked alt-hist early politics to flow faster, give minor amounts of autonomy, and keep you less drained of political power.
- The decisions to modify "Indian Influence" and "Pakistani Influence" now grant autonomy; this was done because a has_autonomy_level check was added down the line for the **Declare Independence** decision, and it bogged down the tree very badly. The changes I did makes it so you don't have to waste as long before breaking away.
- Slightly changed the layout to fix a couple of bugs, improve the layout, and slightly unrailroad the communist path.

**Indian Economy:**
- Fixed one bug with the tree's layout.
- Added line skips throughout the branch and changed the order which certain effects are displayed.
- A few focuses in the economic tree will now improve MIO's.
- Added focus tree path autohiding for the two independent economic paths. Not sure if doing any more could lead to stuff being broken so I left the rest as is.

**Indian Army/Navy/Air**
- Completely rebuilt from scratch, with proper use of MIO's, mastery, special projects, breakthroughs, and every other feature the old tree lacked. Some focuses also give autonomy and there is a small branch related to supporting China.
- There is a unique shared branch between the historical/Unionist-Party to support Britain and her war effort, and a unique shared branch based off the old pro-Soviet one that is now locked down to having any major power ally instead of strictly being the Soviets.


BiccyThiccy:
Before this can be merged, I need to fix some AI stuff. Turns out India never had proper AI strategy plans and most focuses were never given a factor in politics/eco, so me assigning AI_will_do's to my army branches means the AI will only ever do them.

An overview would otherwise be appreciated.

marijn211 (approved):

PR #3361, opened 2026-03-29
Dutch rework splitoff

marijn211:
Very uncontroversial stuff so the origin PR's diff can be shrunk


BiccyThiccy (approved):
Error log is clean and based on some sampling, it seems like the no-MTG deletion was done correctly

PR #3366, opened 2026-04-04
Tiny improvements to Czech

katsuragiyy:
I really wanted to do more but im afraid i just cannot squeeze everything into 7/8th update. Will probably resume after the main update releases.

Adjustments -- >
- Added Czechs to some Check focuses
- Adjusted the DEI flag to be the same as it's overlord counterpart.
- Made it so that Falangism isnt one of the randomly selected ideologies that generated characters can inherit.

GFX -->
- Fixed the Czechoslovak event pictures not showing up.
- Ported over some GFX from the older Czechoslovakia branch.


events/CZE_Czechoslovakia.txt
```diff
@@ -880,7 +880,7 @@ country_event = {
 	id = CZE.203
 	title = CZE.203.t
 	desc = CZE.203.d
-	picture = GFX_CZE_fortification_report
+	picture = GFX_CZKE_fortification_report
```
marijn211: The two fortification event pictures have lowercase CZKE in their GFX tokens in the interface file, not sure if that matters

katsuragiyy: Should probably make them match in order to avoid any missing gfx in the future (again)

marijn211 (changes requested):
Two mentioned event picture tokens should be adjusted (or all the events that use them), everything else is good

events/CZE_Czechoslovakia.txt
```diff
@@ -867,7 +867,7 @@ news_event = {
 	id = CZE.202
 	title = CZE.202.t
 	desc = CZE.202.d
-	picture = GFX_CZE_fortification_news
+	picture = GFX_CZKE_fortification_news
```
marijn211: I think the news variants were lower case as well, please verify

katsuragiyy: They were. Addressed

marijn211 (approved):

PR #3368, opened 2026-04-07
Patch day GER Stuff

BiccyThiccy:


events/r56_Germany.txt
```diff
@@ -15516,4 +15620,73 @@ country_event = {
...
+		NOT = {
+			has_country_leader = {
+				character = GER_august_von_mackensen
+				ruling_only = yes
+			}
+		}
+	}
+
+	mean_time_to_happen = {
+		days = 1825	# Died in 1945
```
marijn211, line 15624: This event does not look configured correctly

e-ee-ee: That's a direct copy past from the vanilla event. I just had to make a new one so it'd get different loc and effects.
Other than that it's identical.

e-ee-ee (approved):
Biccy said I should do this

PR #3373, opened 2026-04-16
Bugfixes + Slight Tweaks for Peace for our Time Release

BiccyThiccy:
**Siam:**
- Fixed a very rare bug (leader switching while an event is active) that could lead to them getting two twin national spirits
- Added a targeted decision to join the factions of other SEA-nation's.
- Added a targeted decision to invite SEA-nations to a faction if they become the leader of one
- Changed one has_country_leader check to has_government check, as the original check was made before I made a second fascist leader option and I forgot to change it.

**Czechoslovakia:**
- Fixed a missing tooltip
- Moved the Czechoslovak Legion focus to be accessible much earlier in the branch. Burying this deep inside one of the mutually-exclusive branches was a huge and unnecessary gameplay hinderance
- Reordered some focuses in the **Static Defense** branch to account for the above change, and to overall make it less railroaded.

**Netherlands:**
- The communist focus to send volunteers is now shared with the fascist branch under a more ideology-neutral name.


common/scripted_triggers/00_scripted_triggers.txt
```diff
@@ -1042,12 +1041,7 @@ is_organization_of_african_unity_member = {
 }
 
 has_tech_sharing_feature = {
-	OR = {
-		# has_dlc = "Together for Victory" # R56
-		# has_dlc = "Death or Dishonor" # R56
-		has_dlc = "Waking the Tiger"
-		has_dlc = "La Resistance"
-	}
+	has_dlc = "La Resistance"
```
marijn211: This is doesn't look equivalent unless it was previously broken, the now always-true check was in an OR, so this would always be true

history/countries/GDC - Guangdong.txt
```diff
@@ -97,9 +97,6 @@ IF = {
 #	add_command_power = 100
 #
 #	if = {
-#		limit = {
-#			has_dlc = "Waking the Tiger"
-#		}
```
marijn211, line 102: If you're breaking if-statements by removing the triggers/breaking bracket structure, its arguably worse than before because it wont work when commented out and might be confusing

common/on_actions/00_on_actions.txt
```diff
@@ -2673,9 +2669,6 @@ on_actions = {
 				}
 			}
 			# RAJ = { # R56
-			# 	if = {
-			# 		limit = { has_dlc = "Together for Victory" }
```
marijn211, line 2677: Idem

common/characters/CAN.txt
```diff
@@ -662,10 +662,6 @@ characters = {
 		#	slot = theorist
 		#	idea_token = CAN_george_volkoff
 		#	ledger = army
-		#	allowed = {
-		#		original_tag = CAN
-		#		has_dlc = "Together for Victory"
-		#	}
```
marijn211, line 668: The first check here in theory still improves performance ever so slightly

marijn211 (changes requested):
[16:26:43][no_game_date][nationalfocus.cpp:625]: Error in focus: MEX_pan_latinism_not_pan_americanism (common/national_focus/mexico.txt:5187 )
relative_focus_id: MEX_focus_march_southwards does not exist. Relative focus must be scripted before this.

marijn211 (approved):
Drei yippies

PR #3375, opened 2026-04-21
Fix non-MtG branches for UK showing

marijn211:
MtG DLC checks were cleaned up but the checks for not having MtG here should've been replaced with always = no.
Currently the DLC and non-DLC versions overlap.


BiccyThiccy (approved):
So vielen jippies! Jippie

PR #3378, opened 2026-04-24
May 2026 Bugfixes

BiccyThiccy:


events/Austria.txt
```diff
@@ -6998,10 +6998,33 @@ country_event = {
...
+							is_core_of = PREV
+							is_on_continent = europe
+						}
+					}
+				}
+				news_event = {
+					id = austria.132
+					days = 1
+					random_hours = 12
+				}
```
marijn211, line 7026: That events seems already set as major = yes, that means this would cause that bug again where everyone gets the same event like 20 times

common/technologies/industry.txt
```diff
@@ -227,6 +232,91 @@ technologies = {
...
+			}
+		}
+		
+		categories = {
+			industry
+			cat_production
+			construction_tech
+			excavation_tech	
+		}
+	}
```
marijn211: Changes to this file should be undone, I already applied these stats to pre-existent tech

events/Austria.txt
ilkober: I copied it from Hotan Clique rebellion r56.news_event.20 - and it seemed only to fire for the countries in question.

You can remove is_major = yes and it should fire correctly

common/national_focus/czechoslovakia.txt
```diff
@@ -2969,9 +2614,6 @@
 			is_in_faction = yes
 			is_faction_leader = no
 		}
-
-		cancel_if_invalid = yes
-		continue_if_invalid = no
```
marijn211, line 2974: Really shouldn't be doing code reorg while the DLC is still getting (relatively) major patches, though I won't hold the PR for it

common/technologies/industry.txt
ilkober: Huh? This is how vanilla handles the energy consumption, what do you want to add instead?

common/ideas/r56_poland.txt
```diff
@@ -1575,7 +1575,7 @@ country = {
 		modifier = {
 			consumer_goods_factor = 0.12
 			production_factory_efficiency_gain_factor = -0.2
-			political_power_cost = 0.6
+			political_power_gain = -0.6
```
marijn211: Why are we removing the stat? Vanilla uses it too

ilkober: Cost is an old stat which equals the negative of gain

marijn211: I understand that but why would we remove it?

common/technologies/industry.txt
marijn211: I already added energy gain to concentrated/distributed industry back when the coal techs were introduced

common/ideas/r56_poland.txt
ilkober: I wanted to make everything coherent, as some trees seem to use both at the same time and it got people confused sometimes

events/Austria.txt
marijn211: Cosmetic feedback: exists = yes is unnecessary in every_other_country scopes

common/national_focus/czechoslovakia.txt
marijn211: Reverted the cleanup commit to make diff comparison easier for the upcoming patches, however we can reapply the commit afterwards

marijn211 (approved):

PR #3386, opened 2026-05-02
British Ensign Flag Fixes + miniscule other flag and country name changes

InfraTheIdiotic:
- Reworked all flags of the British ensign design so that the Union Jack is of the same aspect ratio of the image (therefore fitting into one quarter of the flag as actual British ensign flags do). My method retains the correct shape of the devices in the rest of the flags so they are not stretched.
- Fixed the flag of the Union Jack itself so the left and right sides are not cut off.
- Replaced the Imperial Federation flag with the [British Empire flag](https://en.wikipedia.org/wiki/British_Empire_flag) that was actually contemporarily used, albeit unofficially.
- Replaced the Australian communist flag with a [de facto flag](https://commons.wikimedia.org/wiki/File:A_de_facto_flag_used_by_the_CPA.svg) used by the Australian Communist Party at the time. [Photo](https://archives.anu.edu.au/files/online-exhibition-gallery/images/P126-81-4.JPG)
- Overhauled the Niuean flags to the ones me and Katsura discussed a while back, corrected the Chatham Islands flag, added a communist flag for Tonga and included the British Phosphate Commissioners' logo on the Nauruan fascist flag.
- Changed the communist Irish name to _Irish Workers' Republic_, a name that was popular among Irish communists and can be found used in various 1920s and later sources.
- Changed communist Netherlands Antilles name to Curaçaoan People's Republic, instead of Curative (adjectival form of cure) which was a very weird vanilla name.
- Changed the Newfoundland ENG puppet flags to the red ensign instead of blue as that was more widespread.
- Fixed the Székely Land names so it uses Székely instead of the plural Székelys.


marijn211:
Union Jack seems to have a pretty clear color difference between NZL and AST

InfraTheIdiotic:
Yeah, the flags on Wikipedia/Wikimedia Commons have different colours, they're standardised differently

InfraTheIdiotic:
I personally don't mind it but I can change it to the UK's flag colours if you want, I don't think they were properly standardised like that

marijn211:
> Yeah, the flags on Wikipedia/Wikimedia Commons have different colours, they're standardised differently

I see, interesting

marijn211 (approved):

PR #3387, opened 2026-05-02
New bugfix may

e-ee-ee:


common/decisions/PRC.txt
marijn211, line 4258: Try setting this to any_controlled_state and allowed block to original_tag = JAP, should be more optimized than the current target_root_trigger

marijn211, line 4063: idem, though except the allowed block part

marijn211 (approved):

PR #3388, opened 2026-05-03
Rework Destroyers for Bases to be more historical

marijn211:
(This branch uses files that are only on the bugfix branch, don't try to merge this into the main)


marijn211:
image

PR #3390, opened 2026-05-07
2 bugfixes

e-ee-ee:
[NCNS] Fixed a bug where the RoC would retain a version of the "Army of a Bygone Era" spirit with the DLC tree active.
Fixed several instances where a second German decision regarding a territory could be selected before the first one's effects were applied.
Slightly improved Communist Civil war AI

### DIFF
--- a/common/decisions/r56_GER_decisions.txt
+++ b/common/decisions/r56_GER_decisions.txt
@@ -15412,11 +15412,10 @@ GER_r56_weltpolitik_cat = {
 			base = 10
 		}
 		complete_effect = {
-			set_country_flag = {
-				flag = GER_south_tyrol_active
-				days = 10
-				value = 1
-			}
+			set_country_flag = GER_south_tyrol_active
+		}
+		cancel_effect = {
+			clr_country_flag = GER_south_tyrol_active
 		}
 		remove_effect = {
 			39 = {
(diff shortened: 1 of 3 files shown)


common/decisions/r56_GER_decisions.txt
```diff
@@ -15412,11 +15412,10 @@ GER_r56_weltpolitik_cat = {
...
 		complete_effect = {
-			set_country_flag = {
-				flag = GER_south_tyrol_active
-				days = 10
-				value = 1
-			}
+			set_country_flag = GER_south_tyrol_active
+		}
+		cancel_effect = {
+			clr_country_flag = GER_south_tyrol_active
```
marijn211: Did you test if this works without a cancel_trigger block?

marijn211 (changes requested):
Tested in-game and decisions never seem to cancel

common/decisions/r56_GER_decisions.txt
e-ee-ee: While it doesn't really matter for non-cosmetic purposes, feel free to remove it.
Just don't have time next to studies to do it today and imo keeping it only is a cosmetic issue.

marijn211 (approved):

PR #3397, opened 2026-05-16
Xaelor's June Tweaks n Porting

Xaelor:
Marijn please don't get PTSD while looking at the title...

**Adjustments**
- [GoE] If India gains its independence via its focus tree, it will loose British colonial but gain Indian generals.

**Bugfixes**
- If the Netherlands create the German monarchist puppet, it has some monarchist focuses complete.
- Old RT56 German focus "Increase Influence over the Benelux" now correctly targets Flanders instead of Vladivostok.
- [BftB] The Zveno coup will correctly delete all IMRO militias.

**Adjustments**
- Romanian national spirit "Public Spending" has a smaller consumer goods factor penalty by 15%.

**AI**
- Fixed a bug where the Czechs wouldn't do politics.

**GFX**
- Fixed a few blank generic African portraits.
- Removed Shah Nawaz as a generic Indian portrait.
- [LaR] Fixed a bug when the Ethiopian Fetno Derash intelligence agency didn't have an icon.

**Internal**
- Converted South African legacy characters to the NSB system.
- Removed the "Georgian" (actually Armenian) tree which was hashed out.
- Removed legacy tag aliases for Reichskommissariats.
- Removed some legacy hashed out events.


marijn211 (commented):
>     * Removed some legacy hashed out events.

No need to undo it but please don't do code reorg like this on bugfix PRs

>     * Converted South African legacy characters to the NSB system.
This sorta stuff is okay because the obsolete system causes bugs, however stuff like this

Image
I would be careful with, if South Africa has a civil war and loses the character (not sure if it's possible in practice), this probably throws an error and might even crash the game on focus completion

Xaelor:
> > ```
> > * Removed some legacy hashed out events.
> > ```
> 
> No need to undo it but please don't do code reorg like this on bugfix PRs
> 
> > ```
> > * Converted South African legacy characters to the NSB system.
> > ```
> 
> This sorta stuff is okay because the obsolete system causes bugs, however stuff like this
> 
> Image I would be careful with, if South Africa has a civil war and loses the character (not sure if it's possible in practice), this probably throws an error and might even crash the game on focus completion

It didn't crash @marijn211 but in my opinion it worked poorly (even before). Now it should be better by assigning the traits to random general w/o them and added a proper check in the mission to Europe focus.

marijn211 (approved):

PR #3399, opened 2026-05-24
Harry's Portraits Improvements

Enrx3125:
As the name suggests, this is purely cosmetic. all that was changed was some portraits to improve the quality and also to lessen the use of some from either Kaiserreich or done with AI. 

Some come directly from me while others come from Rise of Nations thanks to the partnership that both mods have, we are able to use those portraits for better quality.


marijn211:
Error log good

marijn211 (approved):

PR #3400, opened 2026-05-25
Biccy June

BiccyThiccy:
Siam:
- Nerfed a couple of overtuned industrial bonuses
- A couple of focuses now modify victory point values
- Buffed most of the diplomacy focuses by either shortening them, improving their decisions or both

Map Changes:
- Gave Siam a new province in its capital state
- Added new victory points throughout the country to hopefully make inland fighting more interesting
- Reworked Panama Canal to be two provinces, with the canal running between the two. This will make it so a nation can invade from north or south (so you don't have to go around the continent under some cases) and allows the US to fight without encirclement penalty.


map/unitstacks.txt
```diff
@@ -258458,7 +258486,7 @@
 8706;38;1523.60;9.50;793.00;-1.57;1.00
 8707;38;2887.28;9.50;720.00;-1.57;0.92
 8708;38;2160.52;9.50;1420.00;-1.57;1.28
-8709;38;2394.38;9.50;1894.00;-1.57;1.28
+8709;38;2394.37;9.50;1894.00;-1.57;1.28
```
marijn211, line 258489: Looks like the nudger changed a lot of unrelated lines in this file? Could you discard those if so.

marijn211 (changes requested):
Above

marijn211 (approved):

PR #3407, opened 2026-06-07
Bugfixing July

katsuragiyy:
do as u will with this 
open to be merged whenever


common/ideas/greece.txt
```diff
@@ -3868,7 +3868,7 @@ GRE_opla = {
 			
 			allowed = {
 				original_tag = GRE
-				has_dlc = "Battle for the Bosporus"
+				has_focus_tree = greek_focus #R56
```
marijn211: Did you test this? Some allowed blocks run before the focus tree selection is done

katsuragiyy: Did just now. It was indeed not a good implementation and caused an error in the error log. Most recent commit should reverse it and use a global flag instead that clears the error (Same issue was present in my tweak to Afghanistan. Implemented the same solution there).

common/ideas/greece.txt
BiccyThiccy, line 1: I'm pretty sure this global_flag won't work with  an allowed check, feel free to discard if you've tested it and found it to work

common/ideas/_manpower.txt
```diff
@@ -436,6 +436,7 @@ ideas = {
 
 			allowed = {
 				original_tag = AFG
+				NOT = { has_global_flag = afg_dlc_focus_tree }
```
BiccyThiccy, line 439: Should be checked also with  both trees

history/countries/INS - Indonesia.txt
BiccyThiccy, line 1: Does any of this intersect with DLC integration changes?

katsuragiyy: Are you referring to the removal of some characters (Github doesn't show what you are referring to)?
If so then no. In Vanilla, the characters i #-out aren't present within the INS - Indonesia file as they are leaders of their respective Sultanates (Aceh, Bali, etc. Whom are all releasables), whom i imported the files of to adjust the capitals and technology to fit with R56.

I haven't played Indonesia yet so im not sure how they can appear in Indo (Since the Sultans have advisor roles that are strictly allowed in INS) but if Vanilla doesn't recruit them at the beginning of the game in the INS history file then i think we're good to go

BiccyThiccy (approved):
PR has six fewer errors than vanilla and diff looks good

PR #3412, opened 2026-06-14
Post taog bugfix part 3?

katsuragiyy:


interface/unitview.gui
marijn211, line 1: It looks like this file wasn't adjusted correctly, firstly because there is still a tab_terrain element which vanilla discarded (which I can see on the vanilla diff repo which I can add you to if you want) and secondly because that is the only diff, which suggests either we shouldn't have this file, or the reason we overwrite this file was wrongly discarded

interface/unitview.gfx
```diff
@@ -170,7 +170,7 @@ spriteTypes = {
 
 	textSpriteType = {
 		name = "GFX_select_all_armylist_short_btn"
-		texturefile = "gfx/interface/select_all_armylist_short_btn.dds"
+		texturefile = "gfx/interface/unitcontrol/select_all_armylist_short_button.dds"
```
marijn211, line 173: This element wasnt changed in 1.18 and 1.19 so that might be an intentional Road edit

common/ai_strategy/JAP.txt
```diff
@@ -298,10 +298,14 @@ southern_expansion = {
 		original_tag = JAP
 	}
 	enable = {
-		has_completed_focus = JAP_nanshin_ron
+		has_completed_focus = has_completed_focus = JAP_strike_the_southern_road
```
marijn211: Typo

interface/unitview.gfx
katsuragiyy: I winmerged the file just incase it contained new changes. This had nothing stating if it was a road change or an outdated vanilla thing, thus resulting in me replacing it (this is me trying to convince you to delete the file ⛓️🕣)

interface/unitview.gui
katsuragiyy: I would really like to not have this file in the mod folder since obviously it adds a lot more to upkeep and also was the cause of multiple crashes. Although I’m also not sure if we have something of value in the file / r56 exclusive that could  harm or break the mod if we were to delete the file

interface/unitview.gfx
marijn211: You can't convince me to delete a file if you don't know what it does 😭

marijn211: Xaelor deleted "unnecessary" interface files before and broke stuff

marijn211: The ideal thing to do is grab the 1.18 version of the vanilla file (which is what we have the private repo for) and run that through a diff checker with the 1.18 Road file, then just add those edits back on top of the 1.19 file, or delete the file if there aren't any we need to keep

katsuragiyy: The .gfx files are used to register graphical assets im pretty sure. In the same manner as ideas.gfx and goals.gfx. So this is a file to register the graphical assets of unit display and such.

marijn211: Indeed, presumably we override it to change the texture of the one of the GFX tokens though

marijn211: We need this file for example to support our 21 unit levels instead of vanilla's 5

marijn211: This is a good edit though, checked in-game

interface/unitview.gui
marijn211: I think Waffle added the file to fix a bug vanilla has now resolved, so in this case can delete

common/ai_strategy/JAP.txt
```diff
@@ -311,21 +315,48 @@ southern_expansion = {
 				has_government = fascism
 			}
 			is_in_faction_with = ENG		#don't hate your allies, pal
+			abort = {
```
marijn211: Typo

marijn211:
[15:20:32][no_game_date][persistent.cpp:67]: Error: "Malformed token: guam_area, near line: 2217" in file: "common/ai_strategy/JAP.txt" near line: 2217
[15:20:32][no_game_date][ai_strategy.cpp:7680]: invalid ai area guam_area
[15:20:32][no_game_date][persistent.cpp:67]: Error: "Unexpected token: area, near line: 2229" in file: "common/ai_strategy/JAP.txt" near line: 2229

Ill fix it myself probs

marijn211 (approved):
Seems good so far, please don't merge, I'll probably do that later

PR #3416, opened 2026-06-17
Performance improvements

marijn211:
Found these with the profiler, please don't overhaul random pieces of code in search of performance.

Currently estimated to cut the hourly ticks by 4.5%.


BiccyThiccy (approved):
Checked diff and error log, super duper cool 😎

PR #3417, opened 2026-06-19
Waffle june 1 - small additions - cleaning errors

Thinking-waffle:
- Added a Hungarian tank scientist.
- Fixing errors caused by bad portrait and advisor pic paths (uppercase, lowercase errors)
- Ported addition of Japanese and British artillery gun models

### DIFF
--- a/common/characters/CAN.txt
+++ b/common/characters/CAN.txt
@@ -300,7 +300,7 @@ characters = {
 		name = CAN_bert_hoffmeister
 		portraits = {
 			army = {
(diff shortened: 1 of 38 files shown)


BiccyThiccy (approved):
Checked the ENG/JAP artillery in-game, and did a sampling of a few of the modified characters, and nothing seems to be messed up.

Can also confirm error log count is lower than beta-edits.

Will do a quick look at the diff and then merge if there's no issues

PR #3419, opened 2026-06-24
[Fix]Lack equipment localisation

zhongtiandang:
I have no idea what these entries with an 'x' mean. It seems like whoever was planning to rework the equipment just forgot about them. We should either fill in the missing locs for them, or actually implement this 'x' equipment system. But at least for now, all the new equipment variants added by Road to 56 are missing localization.


marijn211:
The Road variants are typically like x or a/b/c to distinguish from vanilla, will check PR in a sec

marijn211 (changes requested):
Are you playing with conflicting mods? I dont see this equipment in any Road file

zhongtiandang:
I'm not using any conflicting mods. The reason you guys probably haven't noticed this is that it doesn't seem to occur in the English environment, The problem is this. I only use the Chinese localization, but when I use the equipment added by RT56, it displays this raw code string. I searched the entire mod folder but couldn't find it, so I guess something is broken somewhere..
My guess is that this is legacy content from before the Tank Designer DLC. These entries with an 'x' are likely the old codes from when tank variants had to be researched separately in the tech tree. Since the release of the Tank Designer, this naming issue occurs specifically when designing and naming the RT56 tank variants (like the 1950 modern variant and the 1945 light variant). This problem has been long time.

zhongtiandang:
I think the problem has to do with the DLC compatibility work.

marijn211:
I assumed those names are only used without NSB, lemme check with it enabled

zhongtiandang:
3d73860a-40ce-48a7-a591-ec318041f129
616c08fd-8c20-415b-b38c-9520edb7b79d
61610bdd-fa33-47d8-b97f-db63185a449e
looks like that, it's really not my locas problem.

marijn211 (approved):
Thanks for the clarification

PR #3421, opened 2026-06-26
Shocktroopers subdoctrine track

ilkober:


common/doctrines/subdoctrines/special_forces/special_forces_subdoctrines.txt
```diff
@@ -814,185 +814,265 @@ rangers_2 = {
...
+            	original_tag = SOV
+            	original_tag = GER
+            }          
+        }
+    }
+
+    # EFFECTS
+	effect = {
+		unlock_subunit = shocktroop
+		unlock_subunit = mot_shocktroop
```
marijn211: I think this is incorrect since the other doctrines unlock a support battalion, this already happens (and should happen) on the tech it requires

marijn211 (approved):

PR #3424, opened 2026-06-30
Dutch Selected Additions

DutchPlanDerLindeSus:
This is mostly old stuff containing mainly portraits, gfx, bugfixes and voice changes taken from my earlier work. I didnt see a reason why not to implement this as these are primarily simple changes and fixes including:
- Improved unit voice support for multiple tags
- ~150 new/updated portraits
- Fix to the UK southern shore gfx issue by updating colormap_rgb_cityemissivemask_a
- Improved unit model support and consistency for all tags
- Fix to the persistent "giant horse model" bug when zooming into the map

Launched the game and everything should run fine, no new errors


sound/r56_vo_Thai.asset
marijn211, line 1: Thanks for the smaller PR, I think this file was removed from the mod as it conflicts with the new DLC

marijn211 (commented):
Error log clean, diff seems fine, might take a random sample of portraits to see if they're indeed improvements

marijn211 (commented):
Note to self: Add Quezon portrait switch to PHI election chain, also I randomly remember I think they added an option for a new leader

marijn211 (approved):

PR #3438, opened 2026-07-11
excavation tech AI's tendency BUG

zhongtiandang:
There is an issue with the AI research weights for the Excavation technologies, specifically the extra ones added by RT56. The AI factor is set to 0, which causes the AI to completely ignore late-game excavation techs. Since it's set to 0, it's impossible for the AI to ever pick these techs. I believe this is a bug/oversight because there are still modifiers attached to them.

### DIFF
--- a/common/technologies/industry.txt
+++ b/common/technologies/industry.txt
@@ -1936,7 +1936,7 @@ technologies = {
 		}
 		
 		ai_will_do = {
-			factor = 0
+			factor = 1
 			
 			modifier = {
 				factor = 1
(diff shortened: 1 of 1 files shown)


zhongtiandang:
Maybe the original intention was to set the ai_will_do base to 0 and then add modifiers to increase the weight. But if that's the case, that whole code block would have to be rewritten. Anyway, I'll leave that up to you.

marijn211 (approved):

marijn211:
> Maybe the original intention was to set the ai_will_do base to 0 and then add modifiers to increase the weight. But if that's the case, that whole code block would have to be rewritten. Anyway, I'll leave that up to you.

I've partly done that as it seems logical

PR #3454, opened 2026-08-21
updating displaying sub-ideologies

Flopparanu:


e-ee-ee:
Could you add strasserism to the file as well please?
Didnt realize ypu had to do it back when I added it as an ideology

Flopparanu:
> Could you add strasserism to the file as well please? Didnt realize ypu had to do it back when I added it as an ideology

i've rushed it
i forgot they've added that also

marijn211 (approved):

PR #3456, opened 2026-08-25
FRA localization fix

zhongtiandang:
just localization fix for FRA


localisation/english/FRA_l_english.yml
```diff
@@ -709,7 +709,10 @@
  FRA_commission_fortified_regions_desc:0 "We can recreate the old Commission for Organizing the Fortified Regions of 1927 to study our actual border fortifications and update our defense plans."
  FRA_de_gaulle_strategy:0 "De Gaulle Strategy"
  FRA_de_gaulle_strategy_desc:0 "We cannot sit idly by as enemies move in on our borders. And we cannot leave our allies to fend for themselves in a war scenario. Prepare our army! Gamelin is the man for the job."
+ FRA_motorized_focus:0 "Motorized Focus"
  FRA_motorized_focus_desc:0 "Finding better ways to build our infantry vehicles will help us support and reinforce the front lines."
+ FRA_mechanized_focus:0 "Mechanized Focus"
+ FRA_mechanized_focus_desc:0 "Armored vehicles play an increasingly important role in the war. While slightly less intimidating than your average tank, they give us an edge in communication, command, transport and urban warfare. "
```
marijn211, line 715: These strings load just fine on the release, as they're inherited from vanilla

localisation/english/replace/r56_focus_l_english.yml
marijn211, line 1: All of these are inherited from vanilla where they are just fine, the ones we override generally are improvements on the vanilla loc

marijn211 (changes requested):
All of these lines not being in the mod is intentional

zhongtiandang:
The specific reason why localization is not being inherited from the original version is unclear. According to my tests, it fails to work in non-English languages. I checked the Russian localization mod as well, and it was also ineffective. There could be a strange code error on Paradox's part, but looking into it thoroughly is simply too troublesome.
7b4ceb53118c2f8a47cfdb08c0c397ff
86a21ae4f16303344a23919cce7a4f1e
c8acbfea0bbe07a26132d33fa47076d4
da756387182dcf484f5be889fb7981b9

marijn211:
I might do some duplicate loc cleaning soon that might positively affect this, this solution is not good for the mod

PR #3458, opened 2026-08-28
various bugfixing

e-ee-ee:
- The focus "Weaken the President's Power" has triggers and effects changed when Germany is a constitutional monarchy, allowing for it to still be taken then.
- Fixed a bug where "the fight against antisemitism" national spirit would disappear one day after being added.
- Fixed a bug where Goerdeler would have "Desperate Gambit" after the civil war.
- Fixed a bug where "Institute Price Controls" could be taken after MEFO was cancelled, making the focus useless.
- Fixed a bug where "Beckist Opposition" was not removeable.
- Restricted the focus to restore Brest Litovsk to prevent leftist governments from taking it.
- Prevented an involuntarily subjugated Germany from starting wars.
- Prevented Germany from inviting nations to its faction that are in a civil war.


localisation/english/GER_l_english.yml
```diff
@@ -467,6 +467,7 @@
  GER_wilhelm_ii_2: "Wilhelm II"
  GER_wilhelm_ii_3: "Wilhelm II"
  GER_wilhelm_iv_b: "Wilhelm IV"
+ GER_wilhelm_iv_weak: "Wilhelm IV"
```
marijn211: (Feedback I won't block the PR over) There's a name argument on ideas, more logical to use that than to duplicate the loc

e-ee-ee: should be fixed now

marijn211 (approved):
Since this is stuff you made I'm going to assume its all good, diff looks fine

PR #3467, opened 2026-09-16
Performance Improvements

Xaelor:
- The `any_country` checks have been optimized.

This increases performance by around 3%.
I got this number by playing two games until 1937, one with and the other without the changes and compared the time it took.
Beta edits: 5:53.96
This PR: 5:43.59


Xaelor:
Error log clear @marijn211

marijn211:
This is still code reorg which in the past you botched enough that I've asked you to stop doing, still gonna check this probably since it might be good

marijn211:
Some of the optimized versions have bad tooltips, so there I personally like to stick to alternatives, I can check that though

